Issue #11, opened 2018-01-31
Tampa Bay's too bright


ajbowler:
The Rays color scheme isn't working, the really light blue and the white ain't gonna fly.

Need to test some different colors out from their palette, might have to use a dark font.

PR #19, opened 2018-02-07
Update COL, DET, LAD to more accurate colors.

twfarley:
Added text color to DET, not sure if that works or not.


ajbowler:
Hey, thanks for contributing!

Will test it out on my board later this evening when I get home, just curious where did you get your color values?

I used https://teamcolorcodes.com/mlb-color-codes/ when getting the current ones.

twfarley:
That’s where I got them, too. I noticed COL and LAD were really similar so I went with LAD’s alternate and COL was off a little bit.

Thanks!

ajbowler:
Looks good, particularly the Rockies looking a bit more purple. Thanks!

trevor-viljoen:
@ajbowler Color codes are available through `mlbgame`. I pulled this list together:

https://gist.github.com/trevor-viljoen/6fff77ae591eb788c53961d766ba2f11

ajbowler:
@trevor-viljoen thanks for that, I'll reference this again when the time comes to uplift to mlbgame v3, for now I'm gonna keep using mine since all of the abbrevs are mapped to the abbrev fields in the XML file.

trevor-viljoen:
@ajbowler I used the `team_code` property of the `team` object from the current `mlbgame` release which parses the XML files and not the v3 stuff I've been working on.

Here's they are as RGB colors, if that's more helpful.
https://gist.github.com/trevor-viljoen/7fcf13f187de6d0976e72814e415dd70

ajbowler:
Thanks, I appreciate that.

I'm still not real keen on depending on those JSON keys, though. "chn" translates to Cubs, but the TV scoreboards that I try to mimic with this scorebord show it as "CHC", and it's similar for quite a few other teams as well, mainly for teams that share a city with another one (Chicago, LA, NY) and teams that only have two letters in their abbrev (Giants, Rays, Padres, Royals).

It will require extra code to account for these differences which is why I elected to have my own personal  colors.json, so I can stay as close to the way TV does it. Also allows me to tweak these colors to my liking if something doesn't work quite as well on the board. Just ran into that with #11 a few days ago.

trevor-viljoen:
@ajbowler `display_code` should match up for those. For the Cubs: `chc`, Giants: `sf`, Tampa Bay: `tb`, etc. I went ahead and updated the gists to use display_code instead. Anyway, just thought I'd pass along the info. The project looks cool!

PR #20, opened 2018-02-09
Added the led matrix configuration options to the accepted arguments

swemoney:
This adds the arguments for the rpi-rgb-led-matrix library as passable arguments when starting main.py. Helpful for led matrixes that need a little more configuring (gpio slowdown, more columns than the default 32, brightness, etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,25 @@ See below for more usage options.
 -s, --standings DIVISION  Display standings for the provided division. Example: "NL Central"
 ```
 
+You can configure your LED matrix with the same flags used in the [rpi-rgb-led-matrix](https://github.com/hzeller/rpi-rgb-led-matrix) library. More information on these arguments can be found in the library documentation.
+```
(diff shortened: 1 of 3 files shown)


ajbowler:
Thanks for the contribution, glad to see the adoption of this scoreboard growing!

I'll test this out tonight when I'm off work. I don't use any additional configs for my matrix but this would open up a ton of compatibility with the library.

swemoney:
Sounds great! My pi3 flickers a bit if I don't slow down the GPIO so I figured this would be better than hard coding the options into main. I pulled most of the code directly from the bindings/python/samples code.

ajbowler:
Would you mind updating the Flags section in the README? Just pointing users to the rgbmatrix docs should be fine.

swemoney:
Can do. I'll knock this out a little later today.

swemoney:
Done.

ajbowler:
Merged. Thanks!

PR #24, opened 2018-02-10
Fix innings not fitting on screen when a games goes into extras

swemoney:
Instead of shrinking the inning arrow down to make more space, this fix just aligns the arrow to the top and then slides it over to the left when we hit extra innings to make room for the extra digit.


ajbowler:
Thanks!

I moved the y coordinate for the triangle offset down one more, otherwise the top of the inning would cause it to overflow into the team banner.

Changed that line, then did a rebase/squash. Merged via https://github.com/ajbowler/mlb-led-scoreboard/commit/eca1cc2029a30451768d71c6c3ce93c268c8f5d3

PR #33, opened 2018-02-20
Move all scoreboard related configuration options into a config.json file

swemoney:
Move the commandline arguments that control the scoreboard into a config.json file. led-matrix arguments remain as command line arguments. Also added the config.json to .gitignore.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,6 +1,7 @@
 from renderers.games import GameRenderer
 from rgbmatrix import RGBMatrix, RGBMatrixOptions
 from utils import args, led_matrix_options
+from data.scoreboard_config import ScoreboardConfig
 import renderers.standings
 import renderers.offday
 import datetime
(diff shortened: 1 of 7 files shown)


main.py
```diff
@@ -35,4 +38,4 @@
     else:
       # The mlbgame API returns a 2D array with the list of games as the first index,
       # hence the 'games[0]'
-      GameRenderer(matrix, canvas, games[0], args).render()
+      GameRenderer(matrix, canvas, games[0], args, config).render()
```
ajbowler: Can probably just remove the args param now

utils.py
```diff
@@ -76,3 +70,18 @@ def led_matrix_options(args):
     options.disable_hardware_pulsing = True
 
   return options
+
+class ScoreboardConfig:
```
ajbowler: Move to separate file? Not a big fan personally of having this in utils

ajbowler:
Can you update the README? If not I can do it later today

swemoney:
Should be all set with those changes. Let me know what you think.

Issue #30, opened 2018-02-18
Allow enough time to scroll probable starters across the screen before rotating


ajbowler:
Currently the games (and pregames) rotate every 15 seconds, but for pregames, it may take longer than that to scroll the entire probable starters string across the LED board.

PR #39, opened 2018-02-22
Manually abbreviate team names when the pregame fails to find an overview

swemoney:
Some of these nonsense games aren't returning any overview at all. This probably won't be an issue during the regular season but it's probably a good idea to fail more gracefully anyways. I also append an error message after the pitchers scroll but this will rely on #30 for there to be enough time to actually see the message if rotation is enabled.


ajbowler:
Looks good.

PR #41, opened 2018-02-23
Wait until scrolling has finished once before rotating to the next game

### LINKED ISSUE #30: Allow enough time to scroll probable starters across the screen before rotating

swemoney:
Fixes #30. Waits for scrolling text to complete once before being allowed to rotate to the next pregame.

### DIFF
--- a/renderers/games.py
+++ b/renderers/games.py
@@ -39,6 +39,7 @@ class GameRenderer:
     self.games = games
     self.config = config
     self.current_scrolling_text_pos = self.canvas.width
+    self.current_scrolling_finished_once = False
     self.creation_time = time.time()
 
   def render(self):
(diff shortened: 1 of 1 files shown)


renderers/games.py
```diff
@@ -64,9 +65,10 @@ def render(self):
 
       # TODO: https://github.com/ajbowler/mlb-led-scoreboard/issues/30
       # The time_delta comparison will need to change depending on scrolling text size
-      if self.config.rotate_games and time_delta >= FIFTEEN_SECONDS:
+      if self.config.rotate_games and time_delta >= FIFTEEN_SECONDS and self.current_scrolling_finished_once:
```
ajbowler: Unless I'm missing something this will cause games in progress to never rotate. You'd need to also verify that the game is in a pregame status when checking if the scrolling has finished.

renderers/games.py
```diff
@@ -64,9 +65,10 @@ def render(self):
 
       # TODO: https://github.com/ajbowler/mlb-led-scoreboard/issues/30
```
ajbowler: Mind removing this TODO?

renderers/games.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, matrix, canvas, games, config):
     self.games = games
     self.config = config
     self.current_scrolling_text_pos = self.canvas.width
+    self.current_scrolling_finished_once = False
```
ajbowler: Can you add some docs for this field at the top of the class?

swemoney:
Going to discard this branch for now since we're doing a lot of work in places where this touches. Should be easy to add back in after that work is done.

PR #45, opened 2018-02-24
Split out scoreboard renderer and start work around determining pixel placements depending on matrix size

ajbowler:
Beginnings of moving the pixel numbers out to constants dependent on the provided LED matrix configuration. I think starting down this pattern would help in the long run if 64x32 or other dimensions are to be supported.


ajbowler:
Many changes were made upstream making this effectively a nightmare to rebase so I'll deal with this later. Stability is more important at the moment.

PR #47, opened 2018-02-24
Add basic debug logging

swemoney:
Added a pretty basic debug logger that can be controls via a json attribute in config.json.
Added debug strings to our data objects.


swemoney:
Made a new branch with cleaned up commits.

PR #67, opened 2018-02-28
Stretch the scoreboard to fit 64x32 matrixes

swemoney:
Hits a lot of files but mostly simple stuff. A lot more text centering instead of exact pixels. Everything should automatically stretch across a 64x32 matrix now. Need to make sure I didn't break anything in the process.


renderers/standings.py
```diff
@@ -32,3 +38,26 @@ def render(matrix, canvas, division):
...
+def render_static_wide_standings(matrix, canvas, division, font, text_color):
+  while True:
+    offset = 6
+
+    for team in division.teams:
+      team_text = team.team_abbrev
+      graphics.DrawText(canvas, font, 1, offset, text_color, team_text)
+
+      team_record = str(team.w) + "-" + str(team.l)
+      stat_text = '{:6s} {:4s}'.format(team_record, str(team.gb))
```
ajbowler: Does this handle decimals? It's common to have a team be 3.5 games back or something similar. Should be able to test easily by changing the date passed in main.py

ajbowler:
Would appreciate some added screenshots to `img/` and adding them to the standings section on the README

flemingjw:
The new stretch looks great on my two panels

flemingjw:
Sorry for the sideways photos but the new stretch looks great on my two 32x32 panels

renderers/standings.py
swemoney: Yes. The gb attribute returns - for the division leader and then a 1.0, 2.5, etc for the rest of the teams. I just format it as a string because it's already in the right visual format

PR #111, opened 2018-04-01
Handle manager challenges and delayed start

### LINKED ISSUE #110: Handle Manager Challenges
The data returns a `Manager Challenge` status and currently the game data no longer updates when it enters that status. The Cubs-Marlins game today had a challenge from the Marlins at the end of the 5th that challenged the 3rd out, so it was permanently hung on End 5th.

### LINKED ISSUE #112: Handle "Delayed Start"
If the game is delayed at the start MLB will return `Delayed Start` as the status as opposed to just `Delayed`.

Plan is to just render `Delayed`

ajbowler:
Closes #110
Closes #112 

Also extract statuses out to their own class with some friendly methods

### DIFF
--- a/data/status.py
+++ b/data/status.py
@@ -1,0 +1,45 @@
+class Status:
+  CANCELLED = 'Cancelled'
+  COMPLETED_EARLY = 'Completed Early'
+  DELAYED = 'Delayed'
+  FINAL = 'Final'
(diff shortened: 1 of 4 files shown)


ajbowler:
Giving this one a day or two since I changed all the status logic around to make sure nothing broke

PR #143, opened 2018-04-29
Release 1.4.0

ajbowler:
Merge when milestone is completed and snapshot is dropped from the version string


swemoney:
I'll let you know how I do on finishing the JSON refactor tonight. If all goes well, it might be ok to slip into 1.4.0, otherwise it can wait until 1.5.0

PR #146, opened 2018-05-02
Layouts overhaul

swemoney:
Updated all of the renderers to accept a lot more positional data from the json files so creating alternate layouts and layouts for different sized screens becomes much more trivial.


data/layout.py
```diff
@@ -0,0 +1,75 @@
...
+    self.font_cache = {}
+    default_font = self.__load_font("4x6")
+    self.font_cache = {"4x6": default_font}
+
+  # Returns a dictionary with "font" and "size"
+  def font(self, keypath):
+    d = self.coords(keypath)
+    try:
+      return self.__get_font_object(d[FONTNAME_KEY])
+    except KeyboardInterrupt as e:
```
ajbowler: Why the bailout specifically for the font loading?

renderers/bases.py
```diff
@@ -1,54 +1,46 @@
...
-        if abs(x) == offset and abs(y) == offset:
-          continue
-        self.canvas.SetPixel(base['x'] + x, base['y'] + y, *ledcolors.scoreboard.text)
+        self.__render_baserunner(base_px[base], color)
+
+  def __render_base_outline(self, base, color):
+    x, y = (base["x"], base["y"])
+    size = base["size"]
+    half = abs(size/2)
+    graphics.DrawLine(self.canvas, x + half, y, x, y + half, color)
```
ajbowler: are you shitting me, I didn't know this existed!

utils.py
```diff
@@ -8,11 +9,11 @@ def get_file(path):
 
 def get_font():
```
ajbowler: It looks like this function isn't even being used anymore? Can we remove it?

ajbowler:
Add something to the README for font customization since it's not in the example files by default

PR #148, opened 2018-05-03
Release 1.5.0

ajbowler:
Merge when milestone is completed


ajbowler:
Think we're just waiting on #156 then I'll release

Issue #171, opened 2018-05-23
Display team banners on pregame screen


ajbowler:
Replace the `Team @ Team` text in the pregame screen with the team banners. When the game enters Warmup, replace the game timestamp with the Warmup text.

![image](https://user-images.githubusercontent.com/6600392/40429574-9eb26a84-5e68-11e8-8604-a1c44216eb86.png)

Ignore the win/loss percentages, just display the names.

PR #172, opened 2018-05-26
#171 Display team banners on pregame screen

twfarley:
Changed the pregame status to show the team colors instead of just text.

Needs to be tested during warmup status. It will stop displaying the time and start showing "Warmup" when it changes over.

Probable starter is moved down to match the final and delayed scrolling text positions.

Needs testing against 128 since I don't have access to one.


ajbowler:
@twfarley I get this when running:

```
Traceback (most recent call last):
  File "/home/pi/shared/mlb-led-scoreboard/main.py", line 63, in <module>
    MainRenderer(matrix, data).render()
  File "/home/pi/shared/mlb-led-scoreboard/renderers/main.py", line 38, in render
    self.__draw_game(self.data.current_game(), self.data.overview)
  File "/home/pi/shared/mlb-led-scoreboard/renderers/main.py", line 111, in __draw_game
    renderer = PregameRenderer(self.canvas, pregame, self.data, self.scrolling_text_pos)
TypeError: __init__() takes exactly 6 arguments (5 given)
```

renderers/pregame.py
```diff
@@ -1,38 +1,30 @@
 from data.status import Status
 from rgbmatrix import graphics
 from utils import get_font, center_text_position
+from renderers.teams import TeamsRenderer
 from renderers.scrollingtext import ScrollingText
 import data.layout
 
 class Pregame:
-  def __init__(self, canvas, game, data, probable_starter_pos):
+  def __init__(self, canvas, game, scoreboard, data, probable_starter_pos):
```
ajbowler: The places where this function is called need to be updated to pass in the scoreboard as a parameter

twfarley:
Updated renderers/main.py to include the scoreboard parameter in the pregame renderer. Not sure how that didn't get included in the diff.

Try again?

ajbowler:
Looking good on 32x32 and 64x32. @joelguth could you check 128 when you have some time?

ledcoords/w128h32.json.example
```diff
@@ -126,27 +126,23 @@
...
       "width": 128,
       "warmup": {
         "x": 0,
-        "y": 29,
+        "y": 31,
         "width": 128
       }
     },
     "start_time": {
       "x": 66,
```
joelguth: Change these x coords to 95 and we should be all good for 128x32

PR #177, opened 2018-06-05
Standings coords update

swemoney:
This is a small update to the standings renderer. It breaks up the record and games back into separate coords entries (the `stat` entry on 32x32 boards has been renamed to `record` for conformity sake). On wider screens, `record` is centered along the `x` coord while `games_back` is right aligned. 32x32 screens maintain exact coords since alignment isn't as much of an issue there.

Also redid the fill to take the `height` and `width` for standings into account. The default value for `height` has been changed to `30` to get rid of that annoying extra line of pixels on the bottom of the screen.

### DIFF
--- a/ledcoords/w128h32.json.example
+++ b/ledcoords/w128h32.json.example
@@ -147,7 +147,7 @@
   },
   "standings": {
     "offset": 8,
-    "height": 32,
+    "height": 30,
     "width": 128,
     "divider": {
       "x": 25
@@ -159,8 +159,11 @@
       "name": {
         "x": 1
       },
-      "stat": {
-        "x": 15
+      "record": {
+        "x": 64
+      },
+      "games_back": {
+        "x": 126
       }
     }
   },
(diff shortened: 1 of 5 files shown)


ledcoords/w128h32.json.example
```diff
@@ -159,8 +159,11 @@
       "name": {
         "x": 1
       },
-      "stat": {
```
ajbowler: Are these changes passive to custom coords or is going to break them? I *think* it reads from the default files if it can't find a key in the custom file but I can't remember.

swemoney: This should be pretty passive for custom coords. The `stat` entry in the JSON was renamed so it doesn't exist anymore. Any custom coords that reference `stat` will just be ignored and the new `record` and `games_back` keys will be used instead. So it will require people to update their customizations, but it shouldn't crash or anything.

ajbowler:
@swemoney merge conflicts

I'm good with merging this whenever, if it looks bad on 128 we'll have to wait for someone with 128 to reproduce.

PR #186, opened 2018-07-11
Config file overhaul

swemoney:
I touched a lot of code here but it was mostly all refactoring. Seems to be working well but there could be some issues I missed. Could use some help testing.

This will merge the .example config and the .json config just like the coords and colors so you should only need to add the json keys that you wish to change.

For preferred team/division, I used the plural variants because using multiple teams/divisions for this setting is in the plans so we won't need to change the property name later.


ajbowler:
This is looking great, I'll pull it down within the next week or so. This will go on a 2.0.0 release.

PR #187, opened 2018-07-12
Add rotation option if your preferred team is live but in an inning break

### LINKED ISSUE #185: Allow rotation of games during preferred team's inning break
During an inning break, a user may want to check on how other games are going.

Enhance `stay_on_live_preferred_team` to take in a number.
`0` or `false` means the preferred team can always be rotated over.
`1` or `true` means the preferred team will always be on the board, regardless of inning status.
`2` means the preferred team will be rotated over when it is currently in an inning break.

swemoney:
This should (theoretically) work to allow the rotation to start again when your preferred team is currently playing but the game enters an inning break. This does not check the preferred team again during rotation but it should stop rotation again once the natural rotation hits your preferred team again.

Closes #185


data/status.py
```diff
@@ -46,3 +46,8 @@ def is_fresh(status):
     comes between In Progress and Final and allows a few minutes to see the final outcome before
     the rotation kicks in."""
     return status in [Status.IN_PROGRESS, Status.GAME_OVER]
+
+  @staticmethod
+  def is_inning_break(inning_state):
+    """Returns whether a game is in an inning break (mid/end). Pass in the inning state."""
+    return inning_state in ['Middle','End']
```
ajbowler: These should stored in constants I think

Issue #125, opened 2018-04-20
Enhance rotate_games to use a provided division


ajbowler:
Example: `rotate_games: "NL Central"`

Only rotate through games for teams in the division if that option is provided

PR #188, opened 2018-07-12
Multiple teams and divisions

### LINKED ISSUE #113: Rotate Through Standings
Requested by @swemoney 

Probably need a similar config option to `rotate_games` called `rotate_standings` and have it behave the same, starting on your preferred division if provided.

### LINKED ISSUE #126: Enhance rotate_games to accept an array of team names
If `rotate_games` is given an array, only display the games consisting of those teams.

Example: `rotate_games: ["Nationals", "Orioles", "Mariners", "Brewers"]`

swemoney:
Adds the ability to use an array of strings as the `preferred_teams` and `preferred_divisions` in the config.

`preferred_divisions` will rotate through those divisions when displaying standings.
`preferred_teams` will use the first element as the "favorite" team and will act as it always has before this. A new option in `rotation` called `only_preferred` will limit the rotated games to only games that include teams in the `preferred_teams` list.

Closes #126 
Closes #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ A default `config.json.example` file is included for reference. Copy this file t
 
 ```
 "preferred":                           Options for team and division preference
-  "teams"                      String  Pick a team to display a game for. (This will accept an array in the future) Example: "Cubs"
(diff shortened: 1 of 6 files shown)


ajbowler:
Does this also solve #125? I don't think it does but clarifying anyway

PR #205, opened 2018-12-03
Feature demo

jmorim:
Added demo flag to config with a date parameter to pull data from an example date.


config.json.example
```diff
@@ -25,5 +25,7 @@
 	"end_of_day": "00:00",
 	"full_team_names": true,
 	"scrolling_speed": 2,
-	"debug": false
+	"debug": false,
+	"demo": false,
+	"demo_date": "2017-07-07"
```
ajbowler: I think we should condense this down to one `demo_date` option and have it default to `false`. Then the code can just check if the flag is truthy and parse the date if it's filled out.

data/data.py
```diff
@@ -38,13 +38,18 @@ def __init__(self, config):
     # What game do we want to start on?
     self.current_game_index = self.game_index_for_preferred_team()
     self.current_division_index = 0
+    
+    #
```
ajbowler: remove erroneous comments

PR #207, opened 2018-12-10
Fancy new offday screen

swemoney:
Offday screen updated to display weather and news feed headlines. Bumped version up to 3.0.0 since this requires running the install script again to install the modules needed for RSS parsing and weather fetching from OpenWeatherMaps.

Also, new sections were created in the config so it's probably best to delete your current `config.json` and copy the new `config.json.example` to `config.json` and edit so it fits your needs.


README.md
```diff
@@ -120,6 +130,13 @@ A default `config.json.example` file is included for reference. Copy this file t
     "enabled"                  Bool    Rotation is enabled while your configured preferred_teams game is live.
     "during_inning_breaks"     Bool    Rotation is enabled while your configured preferred_teams game is live during an inning break.
 
+"weather":                             Options for retreiving the weather
```
ajbowler: spelling on retrieving

README.md
```diff
@@ -120,6 +130,13 @@ A default `config.json.example` file is included for reference. Copy this file t
...
+  "apikey"                     String  An API key is requires to use the weather service. You can get one for free at [Open Weather Map](https://home.openweathermap.org/users/sign_up).
+  "zipcode"                    String  The zipcode/postcode for the location you wish to receive weather data
+  "country"                    String  The ISO 3166 country code associated with the zipcode
+  "metric_units"               Bool    Set true for celsius and meters/s. Set false for fahrenheit and miles per hour.
+  "pregame"                    Bool    Add a weather report to the pregame scroll for the game location (Not yet implemented)
```
ajbowler: We should remove config options that aren't implemented. Easy enough to add back in when it actually works

README.md
```diff
@@ -156,6 +173,13 @@ You have the ability to customize the way things are placed on the board (maybe
 ### Custom Colors
 You have the ability to customize the colors of everything on the board. See the `ledcolors/` directory for more information.
 
+### Weather
+This scoreboard will use a weather API to gather weather information at various times. This information is displayed on your teams offdays for your area and also displayed during each game's pregame information. The weather API we use is from OpenWeatherMaps. OpenWeatherMaps API requires an API key to fetch this data so you will need to take a quick minute to sign up for an account and copy your own API key into your `config.json`.
```
ajbowler: We should add a step for this in the installation section since I'm assuming this is required to get the board to run.

swemoney: I added a line/link to the weather section in the installation section. Not having a valid API key fails gracefully and will give a warning in the terminal (it just won't display on the screen if it fails). Let me know if you think we need more here.

PR #216, opened 2019-02-23
Hotfix ajbowler mlbgame

ajbowler:


ajbowler:
Need to update dev branch  to be with master

PR #223, opened 2019-02-24
Exception handling for weather connection issues

swemoney:
We should gracefully fail and try again later when we have a connection issue fetching weather.


data/weather.py
```diff
@@ -54,6 +54,20 @@ def update(self, force=False):
...
+          if self.temp is None:
+            self.temp = -99
+          if self.wind_speed is None:
+            self.wind_speed = -9
+          if self.wind_dir is None:
+            self.wind_dir = 0
+          if self.conditions is None:
+            self.conditions = "Error"
+          if self.icon_name is None:
+            self.icon_name = "50d"
```
ajbowler: I suppose fog is pretty good for when we don't know the weather!

PR #239, opened 2019-03-17
Fix the irregular status games

swemoney:
When I updated all the statuses, I removed everything but the live scoreboard from the `static` definition (since we can scroll text on nearly every game state). This broke our old check (which was already a little broken) so I think this should fix it for good now.


twfarley:
Looks good to me

PR #243, opened 2019-03-18
Use Pythonic Filters

Nixon506E:
The old filters weren't working and I have updated them to use a more pythonic, efficient filter. I saw that the last update fixed the off day filter but used enumerate and didn't do anything to the "only show preferred teams" filter.


ajbowler:
I'm gonna point this to dev so we can test it a bit more with the laundry list of changes we have going for 3.2

PR #253, opened 2019-03-23
Add 24h time format option to config

swemoney:
Added the `"time_format"` option to the config file so you can specify `"12h"` or `"24h"` for your preferred time output format.

Closes #253

### DIFF
--- a/data/scoreboard_config.py
+++ b/data/scoreboard_config.py
@@ -100,9 +100,7 @@ class ScoreboardConfig:
       self.preferred_divisions = [division]
 
   def check_time_format(self):
-    if self.time_format.lower() == "12h":
-      self.time_format = "%-I"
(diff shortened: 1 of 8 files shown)


data/scoreboard_config.py
```diff
@@ -97,6 +99,13 @@ def check_preferred_divisions(self):
       division = self.preferred_divisions
       self.preferred_divisions = [division]
 
+  def check_time_format(self):
+    if self.time_format.lower() == "12h":
+      self.time_format = "%-I"
+    elif self.time_format.lower() == "24h":
```
ajbowler: Would probably be simpler to just check for 24 and set it, otherwise always fall back to 12

swemoney: Ya. That was just silly.

Issue #88, opened 2018-03-13
Handle Doubleheaders


ajbowler:
If a preferred team has two games on a given day, prioritize displaying the game that is currently in progress (first game was played earlier, second game is currently underway).

Currently we show the first of the two games.

demiller9:
My favorite team played a double header today. Neither game was tracked, the scoreboard cycled through the schedule/final all day long.

With some care, you might be able to fix this and handle teams who play successively. (I have a couple of teams in my preferred list. Why can't it follow each game if they are at different times?)

ajbowler:
This issue I actually forgot to close, it should be fixed with the latest version (3.3.0) @demiller9 can you verify 3.3.0 works/doesn't work?

demiller9:
Yes, I am running 3.3.0 (built / installed about two weeks ago).

Issue #51, opened 2018-02-25
Add option to rotate through only live games


swemoney:
A config option that can tell the scoreboard to rotate only through live games. If there are no live games, rotate through all the final scores for the day.

PR #256, opened 2019-03-27
Prioritize live games when there is a multi header

Nixon506E:
#88 Check for multiple games for a team in a given day and then do is_live check when getting current index for a team

### DIFF
--- a/data/data.py
+++ b/data/data.py
@@ -207,10 +207,11 @@ class Data:
     return list(game for game in set(games) if set([game.away_team, game.home_team]).intersection(set(teams)))
 
   def __game_index_for(self, team_name):
-    game_idx = 0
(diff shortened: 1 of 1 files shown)


swemoney:
I really want to do a lot more with detecting live games and doing things differently depending on games that are live but the reason I've always decided not to (yet) is needing to fetch the overview from MLB every time we want to see if the game is live. Some work is being done on moving mlbgame over to the newer json based API which, I believe, we'll have an up to date status for all games in a single update. So this feature (as well as things like #51) have been kind of on the back burner because of it.

I still want to test this out, though, when I have a free moment.

ajbowler:
@swemoney are you still messing with this? This looks good to me as is.

ajbowler:
@swemoney @Nixon506E I am going to merge this into dev. I'm not noticing any slow behavior with the overview refreshes.

PR #257, opened 2019-03-28
Allow bit dithering parameter

Nixon506E:


Nixon506E:
Looks like it currently it is only being used internally by the matrix library but this will at least allow people to use the same options here without error

swemoney:
I'm not sure I see the point in adding an accepted argument that doesn't actually do anything. Just being able to to use the same options seems bad if you're using options that you might expect to change something but it doesn't. Personally, I think it's better to get an error here.

Nixon506E:
We should probably specify that this option is missing in the documentation then as currently it says all options that are supported by the library are supported by the scoreboard.

I wanted to get dithering working but it looks like RGBMatrixOptions doesn't support it.

swemoney:
I agree we need to update the README about that.

ajbowler:
Closing due to inactivity, reopen when changes are made

PR #270, opened 2019-05-15
Add network indicator

swemoney:
- Expanded and enhanced the error logging for network issues
- Catching networking errors in newly merged double header code
- Added a network issues indicator that should show up on the scoreboard when a networking issue is encountered

### DIFF
--- a/data/data.py
+++ b/data/data.py
@@ -17,7 +17,7 @@ try:
 except ImportError:
     from urllib2 import URLError
 
-NETWORK_RETRY_SLEEP_TIME = 5.0
+NETWORK_RETRY_SLEEP_TIME = 10.0
 
 class Data:
   def __init__(self, config):
@@ -41,6 +41,9 @@ class Data:
     self.current_game_index = self.game_index_for_preferred_team()
     self.current_division_index = 0
 
+    # Network status state
+    self.network_issues = False
+
     # Weather info
     self.weather = Weather(self.config.weather_apikey, self.config.weather_location, self.config.weather_metric_units)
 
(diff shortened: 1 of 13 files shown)


data/data.py
```diff
@@ -114,12 +122,17 @@ def refresh_overview(self):
...
         break
       except URLError, e:
+        self.network_issues = True
+        debug.error("Networking Error while refreshing the current overview. {} retries remaining.".format(attempts_remaining))
         debug.error("URLError: {}".format(e.reason))
         attempts_remaining -= 1
         time.sleep(NETWORK_RETRY_SLEEP_TIME)
       except ValueError:
+        self.network_issues = True
+        debug.error("Value Error Occurred while refreshing current game overview. {} retries remaining.".format(attempts_remaining))
```
ajbowler: small thing - this message should have consistent verbiage with the one above

renderers/network.py
```diff
@@ -0,0 +1,29 @@
...
+  def render(self):
+    if self.data.network_issues == True:
+      font = self.layout.font("network")
+      coords = self.layout.coords("network")
+      bg_coords = coords["background"]
+      text_color = self.colors.graphics_color("network.text")
+      bg_color = self.colors.color("network.background")
+
+      # Fill in the background so it's clearly visible
+      for x in range(bg_coords["width"]):
```
ajbowler: can't we use canvas.Fill here? Not sure the difference in performance

renderers/scoreboard.py
```diff
@@ -29,3 +30,4 @@ def render(self):
       PitchesRenderer(self.canvas, self.scoreboard.pitches, self.data).render()
       OutsRenderer(self.canvas, self.scoreboard.outs, self.data).render()
       BasesRenderer(self.canvas, self.scoreboard.bases, self.data).render()
+    NetworkErrorRenderer(self.canvas, self.data).render()
```
ajbowler: Can you move this up to the top of the render function? It's kind of hidden down here at the bottom below the conditional

renderers/network.py
swemoney: canvas.Fill just fills the entire canvas with a color. The indicator is just a small square in the corner and is customizable in the coords so I wouldn't really want to fill the entire screen.

renderers/scoreboard.py
swemoney: I put the network error renderer on the very end of each of the render methods so that it will always appear on top of everything else. The way we do things now, if I put it on top there's a chance other elements will completely cover it up.

Issue #228, opened 2019-02-27
64x32 standings layout not filling in bottom row of LEDs


ajbowler:
![image](https://user-images.githubusercontent.com/6600392/53518099-e7806400-3a95-11e9-9cab-1b4c862413bb.png)

PR #289, opened 2020-07-08
228 standings fix

thalere:
Fixing issue #228 by filling in the last two rows to match the rest of the background.

Also adding some standings render exception logging and the ability to view standings when setting a demo date


ajbowler:
@thalere thanks, released in 3.5.1

Issue #293, opened 2020-07-25
MLB XML API deprecation


ty-porter:
It appears that MLB no longer supports its older GameDay XML API. The API wrapper used by the scoreboard [`mlbgame`](https://github.com/panzarino/mlbgame) depends on this API to render team standings.

There have been lots of reports of the scoreboard warning on startup `Failed to refresh standings.` due to this issue.

A sample API call to retrieve stats:
```
// URL
GET http://mlb.mlb.com/lookup/json/named.historical_standings_schedule_date.bam?season=2020&game_date=%272020/07/25%27&sit_code=%27h0%27&league_id=103&league_id=104&all_star_sw=%27N%27&version=48

// Response
'Check back soon!  Working on you request'
```

Short term, an appropriate fix may be to simply bypass the `mlbgame` dependency/old API only on the broken call and use the new [`MLB Data API`](https://appac.github.io/mlb-data-api-doc). 

Sample call to the new endpoint:

```
// URL
https://statsapi.mlb.com/api/v1/standings?season=2020&leagueId=103,104&date=07/25/2020&division=all

{
  "copyright": "Copyright 2020 MLB Advanced Media, L.P.  Use of any content on this page acknowledges agreement to the terms posted here http://gdx.mlb.com/components/copyright.txt",
  "records": [
    {
      "standingsType": "regularSeason",
      "league": {
        "id": 103,
        "link": "/api/v1/league/103"
      },
      "division": {
        "id": 200,
        "link": "/api/v1/divisions/200"
      },
      "sport": {
        "id": 1,
        "link": "/api/v1/sports/1"
      },
      "lastUpdated": "2020-07-25T18:27:02.000Z",
      "teamRecords": [
        {
          "team": {
            "id": 117,
            "name": "Houston Astros",
            "link": "/api/v1/teams/117"
          },
          "season": "2020",
          "streak": {
            "streakType": "wins",
            "streakNumber": 1,
            "streakCode": "W1"
          },
          "divisionRank": "1",
          "leagueRank": "3",
          "wildCardRank": "3",
          "sportRank": "6",
          "gamesPlayed": 1,
          "gamesBack": "-",
          "wildCardGamesBack": "-",
          "leagueGamesBack": "-",
          "springLeagueGamesBack": "-",
          "sportGamesBack": "0.5",
          "divisionGamesBack": "-",
          "conferenceGamesBack": "-",
          "leagueRecord": {
            "wins": 1,
[...]

PR #294, opened 2020-07-25
293/new standings api

pawptart:
Should partially address [#293] to get standings to display without having to make a major change to how data is handled. This PR bypasses the old XML API to use MLB's new StatsAPI.

There's a new API wrapper designed to mimic the old `mlbgame` version in `data/standings`. Right now it hooks in directly to the old code used so it shouldn't break anything. If the standings fail to render (as they already are), it should fallback to the offday renderer.


pawptart:
Closing this to fix issue with LED driver submodule.

PR #295, opened 2020-07-25
293/fix broken standings

pawptart:
Should partially address [#293] to get standings to display without having to make a major change to how data is handled. This PR bypasses the old XML API to use MLB's new StatsAPI.

There's a new API wrapper designed to mimic the old `mlbgame` version in `data/standings`. Right now it hooks in directly to the old code used so it shouldn't break anything. If the standings fail to render (as they already are), it should fallback to the offday renderer.

### DIFF
--- a/data/data.py
+++ b/data/data.py
@@ -2,8 +2,7 @@ from datetime import datetime, timedelta
 from final import Final
 from pregame import Pregame
 from scoreboard import Scoreboard
-# Import a handler for the standings data to minimize impact to rest of the codebase
-from standings import *
+from standings import Standings, Division, Team
 from status import Status
 from inning import Inning
 from weather import Weather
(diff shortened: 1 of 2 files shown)


data/data.py
```diff
@@ -75,11 +77,17 @@ def set_current_date(self):
...
+    # TODO: This function is a workaround for MLB and mlbgame package no longer supporting GameDay XML API
+    # Old code is left for reference only!
+
+
     try:
-      if self.config.demo_date:
-        self.standings = mlbgame.standings(datetime(self.year, self.month, self.day, 23, 59, 0, 0))
-      else:
-        self.standings = mlbgame.standings()
+      # if self.config.demo_date:
```
swemoney, line 85: Can you get rid of the comments here? The if check appears redundant anyways so just scrap it.

pawptart: Sure thing, taken care of.

data/standings.py
```diff
@@ -0,0 +1,100 @@
+import requests
+import re
+
+class Standings:
+    __URL = 'https://statsapi.mlb.com/api/v1/standings?season={year}&leagueId=103,104&date={month:0>2}/{day:0>2}/{year}&division=all'
```
ajbowler, line 5: Can you extract the `leagueId`'s out to constants? Magic numbers aren't very fun

ajbowler: @pawptart I'm moreso trying to understand what 103/104 actually are. Splitting them out of the main URL string is great but I'm just looking for some human readable context

ajbowler: Is it NL/AL?

pawptart: My bad, I must have been distracted yesterday when I put that up.

Yes, it's just NL/AL.

swemoney: What we'd rather see is something like `LeagueID_NL = 103` and `LeagueID_AL = 104` ... or whichever ID actually means which league. The fact that we don't know is the best reason to break them out into descriptive constants.

pawptart: Totally understand! Should be fixed now.

Issue #299, opened 2020-09-16
feedparser >= 6.0.0 no longer supports Python 2.7


ty-porter:
The latest version of `feedparser` removes support for Python 2.7 used in the scoreboard:

See [this](https://github.com/kurtmckee/feedparser/commit/ce80fcf2582bbf385ec4d6b958ffdff89a8f0480) commit.

The install script for the scoreboard installs the latest version of `feedparser`:

https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/73f2be8d68e2f56aaebfe4f5e75d88e7d3b37d36/install.sh#L16

---

Short term solution is to supply an upper bound to the `feedparser` version.

```bash
sudo pip install "feedparser<6.0.0"
```

However, there is a larger issue with `pip install` under `sudo`, as it can globally overwrite packages and probably isn't desired. Secondly, no packages have locked versions.

Longer term and possibly better solution is to provide a `requirements.txt` and install them with the normal `pip install` command:

```bash
# Someone with a good scoreboard install can run the following and commit it:
pip freeze > requirements.txt

# install.sh
pip install -r requirements.txt
```

pawptart:
Anyone that's already run the installer and is affected by this issue can uninstall the offending `feedparser` version and replace it with the correct one:

```bash
sudo pip uninstall feedparser
sudo pip install "feedparser<6.0.0"
```

PR #301, opened 2020-09-19
Feedparser/python2.7 deprecation

pawptart:
Addresses #299.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+git submodule update
 cd matrix
 echo "Running rgbmatrix installation..."
 sudo apt-get update && sudo apt-get install python2.7-dev python-pillow -y
--- a/install.sh
+++ b/install.sh
@@ -14,7 +14,7 @@ git checkout master
 git fetch origin --prune
 git pull
 sudo apt-get install libxml2-dev libxslt-dev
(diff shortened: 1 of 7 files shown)


install.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+git submodule update
```
ajbowler, line 2: The reason I never had the submodule update in the install script is because users would have different versions of the RBGMatrix library on their systems, making things a bit of a pain to help debug. Keeping that submodule version pinned removes one extra variable.

Any particular reason we need to update the submodule as part of the feedparser fix?

pawptart: I believe `dev` hasn't been updated since the last round of changes a couple months ago. The only change on my commit was `feedparser`: https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/pull/301/commits/7b6da126fe16f8288f1e5b18439177ae4f0ef2bb

ajbowler: I think this change is still coming from your fork, though. I don't see `git submodule update` in master or dev on this repo.

The version bump is definitely from dev, but this change isn't.

pawptart: Hmm. I see that now. I pushed the commit from my Pi I was testing on... Looks like something got out of sync somewhere. I'll close this until I can get a clean PR up.

Issue #300, opened 2020-09-18
ModuleNotFoundError: No module name 'layout'


habeckmatt:
Traceback (most recent call last):
  File "main.py", line 2, in <module>
    from data.scoreboard_config import ScoreboardConfig
  File "/home/pi/mlb-led-scoreboard/data/scoreboard_config.py", line 2, in <module>
    from layout import Layout
ModuleNotFoundError: No module named 'layout'

Any tips on how to resolve this?

pawptart:
This appears to be an issue with compatibility with Python 3.

I currently have a preliminary dev branch going that is aimed at fixing these issues:

https://github.com/pawptart/mlb-led-scoreboard/tree/python3-update

Please note this isn't guaranteed to be bug-free at the moment.

PR #304, opened 2020-09-26
[WIP] Update scoreboard for Python 3 compatibility

ty-porter:
This is a work-in-progress PR to facilitate testing Python 3 compatibility with `mlb-led-scoreboard`. Will address #300 when completed.

Users wishing to upgrade to running the Python 3 upgrade should first check that Python 3.5 or greater is installed:

```bash
python3 --version

# should be 3.5.x or greater
```

Next, checkout this WIP PR branch with the following:

```bash
git fetch origin pull/304/head:python3-update && git checkout python3-update
```

Run the scoreboard installer:

```bash
sh install.sh
```

For existing users, select "No" when prompted to create a new config -- a `config.json` file should already be present and compatible with these changes. **Note**: for test purposes, it would be helpful to set `debug: true` in your config.

Finally, run the board with options:

```bash
# Runs the board and stores the log file for later use
(sudo python3 main.py { YOUR OPTIONS }) %> debug_log.log

# If you want to see the debug text, in a new terminal:
tail -f debug_log.log
```

If you would rather use the `python` instead of `python3` command:

```bash
echo 'alias "python"=python3' >> ~/.bash_aliases
```

---

Once (if?) this PR is merged with `master`, most of these steps will be obsoleted as the `install.sh` script installs from latest `master` branch.

### DIFF
--- a/data/data.py
+++ b/data/data.py
@@ -1,14 +1,14 @@
 from datetime import datetime, timedelta
-from final import Final
-from pregame import Pregame
-from scoreboard import Scoreboard
-from standings import Standings, Division, Team
-from status import Status
-from inning import Inning
-from weather import Weather
-from headlines import Headlines
+from data.final import Final
+from data.pregame import Pregame
+from data.scoreboard import Scoreboard
+from data.standings import Standings, Division, Team
+from data.status import Status
+from data.inning import Inning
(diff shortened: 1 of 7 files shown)


ty-porter:
Closing due to #335

PR #324, opened 2021-04-27
Add support for avoiding scrolling partial messages

ChickenSupreme:
This adds an option to config.json, "avoid_scrolling_partial_messages", which controls whether scrolling text will start scrolling if there's not enough time left in the rotation for the entire message to be displayed. The text will always start scrolling at least once.

I wanted to make this change because with a rotation rate of around 15s and a scrolling speed of 4 many pregame and postgame messages finish scrolling once then start scrolling again but get cut off by the rotation almost immediately. I think it looks better when the scrolling text stays blank after the first scroll until rotating to the next game.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,48 +117,49 @@ See the Flags section below for more flags you can optionally provide.
 A default `config.json.example` file is included for reference. Copy this file to `config.json` and modify the values as needed.
 
 ```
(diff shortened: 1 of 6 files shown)


ChickenSupreme:
I'm going to recreate this PR using the dev branch. Sorry I'm still new to git.

PR #329, opened 2021-05-24
Schema for integration with led-board-manager app

### LINKED ISSUE #330: Schema for config.json added for integration with `led-board-manager` app
I've generated a schema for the config that would allow easy integration to manage boards and generate config forms in the [`led-board-manager`](https://github.com/sflems/led-board-manager) app. This does not affect the core in any way and future features added to this schema would dynamically update the apps forms for config generation/validation and hot-swapping boards.

sflems:
The config.schema.json can be used for validation and also allow for integration with the [`led-board-manager`](https://github.com/sflems/led-board-manager) app to manage scoreboards and auto-generate configuration forms from this schema.

Currently does not affect core. If this schema is updated with future features, users using the GUI app will have a dynamically updated form and validation.

To fix #330

### DIFF
--- a/config.schema.json
+++ b/config.schema.json
@@ -246,7 +246,7 @@
                     "type": "boolean",
                     "title": "avoid_scrolling_partial_messages",
                     "description": "If scrolling text would be interrupted mid scroll by a rotation, don't start scrolling. Text will always be scrolled at least once.",
-                    "default": true
+                    "default": false
                 },
                 "only_preferred": {
                     "$id": "#/properties/rotation/properties/only_preferred",
@@ -316,6 +316,7 @@
             "required": [
                 "enabled",
                 "scroll_until_finished",
+                "avoid_scrolling_partial_messages",
                 "only_preferred",
                 "rates",
                 "while_preferred_team_live"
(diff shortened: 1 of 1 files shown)


ty-porter:
I mean, this seems cool, but why wouldn't this go in the `led-board-manager` repo? Putting this here means someone has to commit to supporting your project.

sflems:
Thanks for the quick response! As for the schema, it's not required for integration, it would simply enhance it. 

This schema file could also be passed through a JSON validator with the config.json at startup, similar to the NHL scoreboard. The error messages returned can then be used to point users in the direction of any invalid data.

If there aren't any core changes planned this would require minimal maintenance and I'd be willing to expand the PR or add another with a validator if you'd like.

No harm no foul if you'd like to exclude it all. 👍 

Cheers!

twfarley:
Tell me a bit more on how this works and we'll consider it. Ty is correct, though. I don't want to have to maintain updates for your project if possible.

If we're just adding a schema that would work fine but it needs to be updated for 5.0.1. Open a new PR if you're still interested, but this sounds like a good idea.

PR #333, opened 2021-05-26
Add support for rotating preferred divisions

### LINKED ISSUE #125: Enhance rotate_games to use a provided division

codydietrich:
This adds support to be able to rotate games through `preferred_divisions` instead of only `preferred_teams` and all games.

This contains a breaking change to the config.json, not sure what the process is on versioning for these. It could probably be modified to be non-breaking if needed.

The dictionary change in `standings.py` is a bit ugly, but I didn't immediately see a way to match the way the game store the team name with the way the standings (and thus the divisions) store the team name.

Closes #125


twfarley:
We can merge this in after fixing the merge conflicts introduced with the recent update. Open a new PR when you're ready. :)

PR #356, opened 2022-04-14
More logos, updated colors, AM/PM in the Pregame

twfarley:


twfarley:
I should mention the colors came from Rich C on the Slack server.

twfarley:
Fixes #357

PR #369, opened 2022-04-25
Truncate team name if Runs or Hits > 9

tracerrx:
If Runs or Hits > 9 (double digits) then truncate the long team name for the display so they don't overflow each other.  There is probably a better way of fixing this.

### DIFF
--- a/renderers/games/teams.py
+++ b/renderers/games/teams.py
@@ -82,7 +82,7 @@ def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, 
     coords = layout.coords("teams.name.{}".format(homeaway))
     font = layout.font("teams.name.{}".format(homeaway))
     team_text = "{:3s}".format(team.abbrev.upper())
-    if full_team_names and canvas.width > 32 and not (short_team_names_for_runs_hits and (team.runs > 9 or team.hits > 9 or vs_team.runs > 9 or vs_team.hits > 9)):
+    if use_full_team_names:
         team_text = "{:13s}".format(team.name)
     graphics.DrawText(canvas, font["font"], coords["x"], coords["y"], text_color_graphic, team_text)
 
--- a/renderers/games/teams.py
+++ b/renderers/games/teams.py
@@ -53,8 +53,8 @@ def render_team_banner(canvas, layout, team_colors, home_team, away_team, full_t
                 y_offset = accent_coords[team]["y"]
                 canvas.SetPixel(x + x_offset, y + y_offset, color["r"], color["g"], color["b"])
 
-    __render_team_text(canvas, layout, away_colors, away_team, "away", full_team_names, default_colors, short_team_names_for_runs_hits)
-    __render_team_text(canvas, layout, home_colors, home_team, "home", full_team_names, default_colors, short_team_names_for_runs_hits)
+    __render_team_text(canvas, layout, away_colors, away_team, "away", full_team_names, default_colors, short_team_names_for_runs_hits, home_team)
+    __render_team_text(canvas, layout, home_colors, home_team, "home", full_team_names, default_colors, short_team_names_for_runs_hits, away_team)
 
     # Number of characters in each score.
     score_spacing = {
@@ -74,13 +74,13 @@ def __team_colors(team_colors, team_abbrev):
     return team_colors
 
 
-def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, default_colors, short_team_names_for_runs_hits):
+def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, default_colors, short_team_names_for_runs_hits, vs_team):
     text_color = colors.get("text", default_colors["text"])
     text_color_graphic = graphics.Color(text_color["r"], text_color["g"], text_color["b"])
     coords = layout.coords("teams.name.{}".format(homeaway))
     font = layout.font("teams.name.{}".format(homeaway))
     team_text = "{:3s}".format(team.abbrev.upper())
-    if full_team_names and canvas.width > 32 and not (short_team_names_for_runs_hits and (team.runs > 9 or team.hits > 9)):
+    if full_team_names and canvas.width > 32 and not (short_team_names_for_runs_hits and (team.runs > 9 or team.hits > 9 or vs_team.runs > 9 or vs_team.hits > 9)):
         team_text = "{:13s}".format(team.name)
     graphics.DrawText(canvas, font["font"], coords["x"], coords["y"], text_color_graphic, team_text)
 
--- a/renderers/games/teams.py
+++ b/renderers/games/teams.py
@@ -86,6 +86,15 @@ def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, 
         team_text = "{:13s}".format(team.name)
     graphics.DrawText(canvas, font["font"], coords["x"], coords["y"], text_color_graphic, team_text)
 
+def __can_use_full_team_names(canvas, enabled, abbreviate_on_overflow, teams):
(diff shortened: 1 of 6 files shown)


WardBrian:
This logic should probably be in the display (renderers/games/teams.py), not inside data/game.py. It can also be much shorter if you just use `self.away_hits()` rather than re-creating the dict indexing.

I think it would also be better if it switched to using the abbreviations (like the short name setting) rather than just truncating, IMO

tracerrx:
Yes this occurs on the 128 x 64 layout when RH > 9 and you have team names like GUARDIANS.

twfarley:
If I have errors and hits off will the scores still be truncated?

WardBrian:
> If I have errors and hits off will the scores still be truncated?

This is another good reason to move this logic into the renderer, since it has access to that config in the layout

tracerrx:
Ok, will keep working on it... Which settings file should I use the config.json or where we enable RHE?

tracerrx:
Took your guidance and added a setting to config.json (short_team_names_for_runs_hits), and moved the logic to render.  However now it seems to always choose the abrev team name...

WardBrian:
I think it should probably go alongside the config for RHE themselves in the layouts. As for why it's always on, I think adding the `and short_team_names_for_runs_hits == False` is responsible?

tracerrx:
Actually what i want is, full team names normally, unless runs OR hits are > 9, then return Abrev team name.

WardBrian:
Yes, but the existing logic is the flip of that. By default it assigns it to the abbreviation a few lines above these if statements, which is probably the right thing since we want to force it to be abbreviations on 32x32

tracerrx:
image

pjockey:
10 hits!

pjockey:
10 hits, full names

tracerrx:
@pjockey did you have `"short_team_names_for_runs_hits": true,` in your config.json?

pjockey:
> @pjockey did you have `"short_team_names_for_runs_hits": true,` in your config.json?

no, nothing like that in the comfig.json.

tracerrx:
@pjockey add it to your config.json after the line that says `"full_team_names": true,`  then STOP and START the board to re-load the config.

pjockey:
Done,. 
You appear to be using a different font than what I am. Look at the zeros.

tracerrx:
@pjockey im using a 128x64 2mm pitch... And I did change the font and some other things in the layout file...

pjockey:
> @pjockey im using a 128x64 2mm pitch... And I did change the font and some other things in the layout file...

That's what I thought, 128x64. I'm on 64x32. need to upgrade. I do have another 64x32 board  on hand. Should attach them.

tracerrx:
May change it to truncate both teams names. Screen Shot 2022-04-26 at 9 38 59 PM
Screen Shot 2022-04-26 at 9 38 50 PM

WardBrian:
I think truncating both makes the most sense

tracerrx:
Going to have to rework some stuff since were not passing both home and away.. only team... Unless you can think of another way

WardBrian:
I’d do it by moving both render_team calls into a new function which first determines which size name to do and passes that as an argument to both

tracerrx:
Works as expected 
image

pjockey:
Looks good. Where did you get your board from? And, which branch of this build has all of the recent changes?  I have @WardBrian 's fork running. I says it is up to date, yet I have no backwards "K", etc.

WardBrian:
My fork was merged into the main branch here

pjockey:
> My fork was merged into the main branch here

Is there a git command to merge or should I just re-install and start afresh?

renderers/games/teams.py
```diff
@@ -53,8 +53,8 @@ def render_team_banner(canvas, layout, team_colors, home_team, away_team, full_t
                 y_offset = accent_coords[team]["y"]
                 canvas.SetPixel(x + x_offset, y + y_offset, color["r"], color["g"], color["b"])
 
-    __render_team_text(canvas, layout, away_colors, away_team, "away", full_team_names, default_colors)
-    __render_team_text(canvas, layout, home_colors, home_team, "home", full_team_names, default_colors)
+    __render_team_text(canvas, layout, away_colors, away_team, "away", full_team_names, default_colors, short_team_names_for_runs_hits, home_team)
+    __render_team_text(canvas, layout, home_colors, home_team, "home", full_team_names, default_colors, short_team_names_for_runs_hits, away_team)
```
ty-porter, line 57: This arg list is getting out of hand, can we create a helper function:

```python
def can_use_full_team_names(canvas, enabled, abbreviate_on_overflow, teams):
  if enabled and canvas.width > 32:

    if abbreviate_on_overflow:
      for team in teams:
        if team.runs > 9 or team.hits > 9:
          return False
    else:
      return True
  
  return False
```

This would let us do this:
```suggestion
    use_full_team_names = can_use_full_team_names(canvas, full_team_names, short_team_names_for_runs_hits, [home_team, away_team])

    __render_team_text(canvas, layout, away_colors, away_team, "away", use_full_team_names, default_colors)
    __render_team_text(canvas, layout, home_colors, home_team, "home", use_full_team_names, default_colors)
```

See next comment for simplifying line 83

renderers/games/teams.py
```diff
@@ -74,13 +74,13 @@ def __team_colors(team_colors, team_abbrev):
...
 
-def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, default_colors):
+def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, default_colors, short_team_names_for_runs_hits, vs_team):
     text_color = colors.get("text", default_colors["text"])
     text_color_graphic = graphics.Color(text_color["r"], text_color["g"], text_color["b"])
     coords = layout.coords("teams.name.{}".format(homeaway))
     font = layout.font("teams.name.{}".format(homeaway))
     team_text = "{:3s}".format(team.abbrev.upper())
-    if full_team_names and canvas.width > 32:
+    if full_team_names and canvas.width > 32 and not (short_team_names_for_runs_hits and (team.runs > 9 or team.hits > 9 or vs_team.runs > 9 or vs_team.hits > 9)):
```
ty-porter, line 83:
```suggestion
    if use_full_team_names:
```

ty-porter (changes requested):
The logic looks good!

What do you think about making these changes? Our argument list for `__render_team_text` is huge and unorganized, this would fix that

renderers/games/teams.py
```diff
@@ -53,8 +53,10 @@ def render_team_banner(canvas, layout, team_colors, home_team, away_team, full_t
                 y_offset = accent_coords[team]["y"]
                 canvas.SetPixel(x + x_offset, y + y_offset, color["r"], color["g"], color["b"])
 
-    __render_team_text(canvas, layout, away_colors, away_team, "away", full_team_names, default_colors)
-    __render_team_text(canvas, layout, home_colors, home_team, "home", full_team_names, default_colors)
+    use_full_team_names = can_use_full_team_names(canvas, full_team_names, short_team_names_for_runs_hits, [home_team, away_team])
```
ty-porter, line 56: This method definition is missing I think

tracerrx: Not sure I can do this on my side, Sorry.. New to using GIT and think I merged locally which messed up my branches/PR...```
 > git pull
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=<remote>/<branch> fix-teamname-rhe-display
```

ty-porter (approved):

PR #370, opened 2022-04-26
Add pregame weather

tracerrx:
Added pregame weather.  In config.json must add `"pregame_weather": true,` note that pregame weather appears to be entered manually when the booth crew shows up.  For tonight's yankees game the api populated around 3pm today for a 7:05 start.

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -43,6 +43,7 @@
 	"end_of_day": "00:00",
 	"full_team_names": true,
 	"short_team_names_for_runs_hits": true,
+	"pregame_weather": true,
 	"scrolling_speed": 2,
 	"debug": false,
 	"demo_date": false
--- a/data/config/__init__.py
+++ b/data/config/__init__.py
@@ -63,6 +63,7 @@ class Config:
         self.end_of_day = json["end_of_day"]
         self.full_team_names = json["full_team_names"]
         self.short_team_names_for_runs_hits = json["short_team_names_for_runs_hits"]
+        self.pregame_weather = json["pregame_weather"]
         self.debug = json["debug"]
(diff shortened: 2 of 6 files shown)


tracerrx:
Screen Shot 2022-04-26 at 3 14 40 PM
Screen Shot 2022-04-26 at 3 14 28 PM
Screen Shot 2022-04-26 at 3 14 16 PM

twfarley:
That’s cool. I wonder if it makes sense to put it on the same line as the start time. Thoughts?

WardBrian:
You could have the setting be like debug_day where it’s either False, “time”, or “starters”

tracerrx:
> That’s cool. I wonder if it makes sense to put it on the same line as the start time. Thoughts?

I figured we would put the National Broadcast Station on the game time line in the future

tracerrx:
Yankees game weather just changed to "Drizzle"... Cool... Nationals/Marlins weather also just changed, so someone is updating them.

twfarley:
Do we get the name of the park where the game is played? If so, we should put this on the bottom line with the start time like this:

7:05PM - Yankee Stadium - 65F Drizzle

tracerrx:
Look like "officialVenue" is an link to another api call... So its possible but would require an additional api call I believe.  @WardBrian can clarify.

tracerrx:
Scratch that.. looks like it under both? Screen Shot 2022-04-26 at 7 04 15 PM Wonder what the difference is between venue and officialVenue?

ty-porter (approved):
Unclear why the diff didn't update before swapping branches, but this LGTM now

PR #375, opened 2022-04-29
Pitch data

tracerrx:


tracerrx:
We still need to integrate the pitch data into other size boards

WardBrian:
I think it will be hard to fit on anything smaller, so you might have to do what I did with the at bat information and either put it at a coordinate outside the region of the board or disable it if there’s a setting to do so

tracerrx:
I'm sure someone can manage to squeeze in the mph  and maybe even short pitch type on some of the displays, but for now I defaulted them all to false except for the 128x64

ty-porter:
@tracerrx Can you fix the conflict for the field names and I'll get this merged to dev

tracerrx:
@ty-porter can you just double check that, not in a spot where I can actually run it

PR #377, opened 2022-05-03
Fix double logger in RGBME

ty-porter:


WardBrian (approved):
Confirmed this fixed the issue for me

PR #381, opened 2022-05-15
Tweak: Left justify batter and pitcher names if they are short enough to avoid scrolling

WardBrian:
We could make this a config in the layout if we wanted, for now I'm assuming it is better default behavior.

Comparison (specifically noticeable for short batter names like Soto):

Before:
before

After:
after


WardBrian:
@ty-porter was this closed on purpose or as a side effect of the release process?

ty-porter:
Sorry about that, it was closed during release when I spun up a new dev branch and killed off the old one.

WardBrian:
@twfarley this PR now hides the scores on the pregame in addition to justifying the pitcher/batter

twfarley:
looks good to me. thanks, Brian.

PR #392, opened 2022-07-09
Update to fix issue with python 3.10+ and Mapping library.

brentm5:
Due to a deprecated function collections.Mapping no longer works.
Instead you need to import mapping with collections.abc

I ran into the following issue trying to rebuild my scoreboard. 

```
ERROR (02:39:02): Untrapped error in main!
Traceback (most recent call last):
  File "/opt/mlb-scoreboard/main.py", line 141, in <module>
    main(matrix, config)
  File "/opt/mlb-scoreboard/main.py", line 34, in main
    config = Config(config_base, matrix.width, matrix.height)
  File "/opt/mlb-scoreboard/data/config/__init__.py", line 22, in __init__
    json = self.__get_config(filename_base)
  File "/opt/mlb-scoreboard/data/config/__init__.py", line 193, in __get_config
    new_config = deep_update(reference_config, custom_config)
  File "/opt/mlb-scoreboard/utils.py", line 183, in deep_update
    if isinstance(value, collections.Mapping) and value:
AttributeError: module 'collections' has no attribute 'Mapping'
```

From looking it appears `collections.Mapping` no longer exists in python 3.10+ so I updated it to use the new `collections.abc` recommendation.  From my quick reading this should work all the way back to python 3.3.


WardBrian:
Seems like a whitespace issue (add ?w=1 to the diff URL and it goes away). Maybe spaces -> tabs?

brentm5:
Let me fix that. I must have missed that with my local diff.

brentm5:
I also added an editor config which should help with this happening in the future.  Either git (via core.autocrlf true in git) or my editor saved it as this instead of this instead of the original.

PR #407, opened 2022-09-01
Added Pitch Count

tracerrx:
Added pitch count to the pitcher scroll.  Set show_pitch_count to true/false in config.json to enable/disable

### DIFF
--- a/renderers/games/game.py
+++ b/renderers/games/game.py
@@ -3,6 +3,7 @@ try:
 except ImportError:
     from RGBMatrixEmulator import graphics
 
+from distutils import config
 from data.config.color import Color
 from data.config.layout import Layout
 from data.scoreboard import Scoreboard
@@ -14,7 +15,7 @@ from renderers import scrollingtext
 from renderers.games import nohitter
 
 
-def render_live_game(canvas, layout: Layout, colors: Color, scoreboard: Scoreboard, text_pos, animation_time):
+def render_live_game(canvas, layout: Layout, colors: Color, scoreboard: Scoreboard, text_pos, animation_time, config):
     pos = 0
 
     if scoreboard.inning.state == Inning.TOP or scoreboard.inning.state == Inning.BOTTOM:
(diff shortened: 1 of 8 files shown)


ty-porter:
Agreed, `Game` objects are data containers. Turning config on or off should swap display options, not what the underlying data includes (unless there's a good reason for that)

WardBrian:
Storing it alongside balls/strikes/pitch speed also allows easier control over where the user would like it to be. For example, I think I'd want it as a separate text element rather than part of the pitcher's name

tracerrx:
I think this moves it to the right place

WardBrian:
How do you feel about this setting being in the layout rather than the config?

ty-porter:
Correct me if I'm wrong... The reason it's a config setting is because we want it to scroll with the existing pitcher text?

So our options are:
* layout (static text)

WardBrian:
I wouldn’t want it on the pitcher text I think, I would want it like the speeds (probably under them, on my layout), but it could be an “off” “pitcher name” “separate” setting in the layout

renderers/games/game.py
```diff
@@ -3,6 +3,7 @@
 except ImportError:
     from RGBMatrixEmulator import graphics
 
+from distutils import config
```
ty-porter, line 6: This import is not needed, probably breaks some stuff

tracerrx:
I understand the logic of having it an independent and being able to move it around the layout... But wouldn't that prohibit it from being part of the pitcher scroll?

WardBrian:
I don’t think so, you should be able to append it right before rendering. 

I can sketch this out in my own fork if it’s helpful, you already did the hard part of finding where the data lives in the API

tracerrx:
So in layout file, under atbat something like `"pitch_count": {
      "font_name": "4x6",
      "x": 1,
      "y": 50,
      "enabled": true,
      "append_pitcher_Name": true
    },`

And if append pitch name==true, don't render separately but append to the pitcher name.

WardBrian:
That sounds good to me

tracerrx:
Sure.. I prob can’t get around to it until Tuesday though

WardBrian:
#431 implemented this feature in a way that only uses the coordinates file, not the config. You can see the example here: 
https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/0fee496dfa68232ab07daba329f899f70aeada50/coordinates/w64h32.json.example#L135-L140

PR #412, opened 2022-10-03
Fix playoff bracket

WardBrian:
This changes the playoff bracket to properly render the new wild card games. 

screenshot


pjockey:
Is there a tweak to have the brackets show?

WardBrian:
They should automatically replace the standings when the postseason starts. If you want to preview them now like in the screenshot you can set the debug date to `"2022-10-07"`

Issue #422, opened 2023-02-25
Automatically update custom configurations on install


ty-porter:
Adding new config options is pretty confusing for end users, the existing messaging leads users to completely regenerate the config. However the config is pretty big, so updating it by hand tends not to happen, and if there are new keys then they will not be in the config and can lead to issues, such as (possibly) #387

When **updating** an existing installation that has a custom configuration (`config.json`)

1. Automatically delete any keys that are no longer used (if any)
2. Merge any missing keys with default values
3. Provide a list of changes as feedback to the user once install is completed

If installing fresh, the existing copy from `config.json.example` -> `config.json` is still appropriate

PR #423, opened 2023-02-25
Upsert missing config keys on install or update

ty-porter:
Addresses #422


WardBrian (approved):
This looks great.

We should consider doing a lot more testing for things we can. The first that came to mind is it would be great to have a test which verifies that all the team names/abbreviations match what we have stored (e.g. to prevent #424), or to test that all the news feed links actually resolve to something still

PR #424, opened 2023-02-28
 Fix: D-backs changed short code, ARI -> AZ 

WardBrian:


WardBrian:
Oops. wasn't very clean with my git history so this also includes #418

WardBrian:
#423 would be very good for this I think

ty-porter (approved):

PR #427, opened 2023-03-02
Deprecate rotation rates as single Float

ty-porter:
With the new config updater, there's a chance to overwrite this value with schema defaults. This PR formally deprecates that functionality and provides an upgrade path to the full dict format.

Deprecates:
```json

{
  "rotation": {
    "rates": 15.0
  }
}
```

Converts to:
```json
{
  "rotation": {
    "rates": {
	"live": 15.0,
	"final": 15.0,
	"pregame": 15.0
    }
}
```


data/config/__init__.py
```diff
@@ -179,7 +183,7 @@ def read_json(self, path):
         if os.path.isfile(path):
             j = json.load(open(path))
         else:
-            debug.warning(f"Could not find json file {path}.  Skipping.")
+            debug.info(f"Could not find json file {path}.  Skipping.")
```
ty-porter, line 186: This really throws people off this was a WARN instead of INFO

data/config/__init__.py
```diff
@@ -128,6 +128,10 @@ def check_rotate_rates(self):
             try:
                 rate = float(self.rotation_rates)
                 self.rotation_rates = {"live": rate, "final": rate, "pregame": rate}
+
+                debug.warning("(DEPRECATION) Config option rotation->rates no longer supports single Float values!")
+                debug.warning("              Re-run the install/update script to convert to the new format")
+                debug.warning("              Example: sudo sh ./install.sh")
             except:
                 debug.warning(
                     "rotation_rates should be a Dict or Float. Using default value. {}".format(DEFAULT_ROTATE_RATES)
```
WardBrian, line 137: We should remove the 'or Float' here

Issue #416, opened 2022-10-13
Awkward letter spacing


Bigred200:
after the use of a single column letter or number most commonly seen in the use of "i or l" there is an extra space between the next letter. Example  "framber val dez"  had to leave before I could try to adjust anything so I figured I'd post and see if there was a quick fix I could try before I just started guessing once I get home. Thanks

WardBrian:
I think this came up before and was specific to some version of the matrix library. Smaller letters like that should be centered

Bigred200:
What should I try?

ty-porter:
Certain versions of the matrix driver library left-aligned characters. I haven't followed lately to know if it's been fixed. The solution is to checkout a known good version of the driver lib and reinstall.

https://mlb-led-scoreboard.slack.com/archives/C9BLJ9JF9/p1651458034419249?thread_ts=1651432875.074229&cid=C9BLJ9JF9

c2mfj:
> 
I installed just a few weeks ago and am experiencing this, I do not have access to slack, how can I get an invite?

ty-porter:
Click the [![Join Slack](https://img.shields.io/badge/slack-join-blue.svg)](https://join.slack.com/t/mlb-led-scoreboard/shared_invite/zt-1f6n4axo4-r32OH7dlSAjEjstFV4RDNQ) button here or at the top of the README

Just to put the info from the Slack thread here so we don't lose it:

The initial recommendation 10 months ago was revert to a previous commit of the driver library
```
# from scoreboard dir
cd submodules/matrix

sudo git checkout a56338db0f003d5236f2ce98c73a591d64a70852
```

I don't recommend that any more as that commit is over a year old at this point

However, while I was testing this, our patched 4x6 font doesn't match the matrix driver 4x6 font
https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/fonts/4x6.bdf#L996-L1007
https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/master/assets/fonts/patched/4x6.bdf

I think I patched the wrong font...

ty-porter:
All this to say:

I think there's a bug in `rpi-rgb-led-matrix` that left-aligns glyphs. BDF fonts have settings per-character such as bounding boxes, glyph offsets, etc., that are being ignored:

[...]

PR #428, opened 2023-03-03
Fix misaligned characters in 4x6 font

ty-porter:
Addresses #416


WardBrian:
Is this font visually distinct from the one we currently use, besides that issue being fixed?

I ask because my pi is still running a rather old RGB matrix driver version which doesn’t yet have that issue

ty-porter:
Yes, this is unfortunately a "breaking" change. Our 4x6 font was the original `tom-thumb` font, so this reverts to that.

I added a GIF of common characters to the PR

WardBrian:
Thanks! Very helpful. If not for that capital N I’d say this is a straight up improvement.

WardBrian:
Could we keep a (renamed) version of the current font in the repo such that people could still use it in layouts?

ty-porter:
Yeah! The great thing about it is you can always revert back to `assets/fonts/patched/tom-thumb.bdf` instead.

WardBrian (approved):

PR #430, opened 2023-03-11
Add series status to pre/postgame scroll during playoffs

WardBrian:
Based on a PR I did for MLB-StatsAPI last year: https://github.com/toddrob99/MLB-StatsAPI/pull/95

Adds text like "HOU leads 2-0", or "Series tied 1-1" to the end of the scroll (after the weather for pregame, after the save pitcher for postgame).

Requires MLB-StatsAPI version 1.6 or higher

### DIFF
--- a/data/game.py
+++ b/data/game.py
@@ -1,5 +1,6 @@
 import time
 from datetime import datetime
+from typing import Optional
 
 import statsapi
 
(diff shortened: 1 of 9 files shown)


WardBrian:
Here's a example from last year's World Series:

example

ty-porter (approved):
LGTM

PR #431, opened 2023-03-12
Add ability to show pitch counts

WardBrian:
Updated version of #407 

My plan is to use this as a replacement for the pitch speed during the no hitter/ perfect game layout

Fake data demo:
demo

Can also be appended to a pitcher's name in the coordinates file:

demo 2


coordinates/w64h32.json.example
```diff
@@ -126,14 +126,18 @@
...
-      "font_name": "4x6",
       "x": 1,
       "y": 50,
-      "width": 41,
       "enabled": false,
       "mph": false,
       "desc_length": "Short"
     },
+    "pitch_count": {
+      "x": 1,
```
ty-porter, line 136: This one looks like it's missing `font_name`

ty-porter (approved):

coordinates/w64h32.json.example
WardBrian: Yeah, I deleted it there since 4x6 is the default for this size. It’s not clear to me if we even should be setting it on the other boards in the examples

Issue #432, opened 2023-03-12
Testing: Verify all existing coordinate examples are valid


WardBrian:
"Valid" in the sense that they meet the schema, not that they're visually correct.

PR #437, opened 2023-03-25
Verifying all coordinates are valid against master schema

ty-porter:
Addresses #432

### DIFF
--- a/coordinates/w128h32.json.example
+++ b/coordinates/w128h32.json.example
@@ -209,8 +209,8 @@
     }
   },
   "standings": {
-    "__comment__": "Delete this next line and make offset 8 to make the font much larger but only display 4 teams",
     "font_name": "4x6",
+    "start": 0,
     "offset": 6,
     "height": 32,
     "width": 128,
--- a/coordinates/w128h64.json.example
+++ b/coordinates/w128h64.json.example
@@ -188,6 +188,7 @@
     }
   },
   "standings": {
+    "font_name": "4x6",
     "start": 1,
     "offset": 12,
     "height": 60,
--- a/coordinates/w32h32.json.example
+++ b/coordinates/w32h32.json.example
@@ -206,6 +206,8 @@
     }
   },
   "standings": {
+    "font_name": "4x6",
+    "start": 0,
     "offset": 6,
     "height": 30,
     "width": 32,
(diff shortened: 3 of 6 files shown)


coordinates/w128h32.json.example
```diff
@@ -209,8 +209,8 @@
     }
   },
   "standings": {
-    "__comment__": "Delete this next line and make offset 8 to make the font much larger but only display 4 teams",
     "font_name": "4x6",
+    "start": 0,
```
ty-porter, line 213: This defaults to 0 in code

coordinates/w128h64.json.example
```diff
@@ -188,6 +188,7 @@
     }
   },
   "standings": {
+    "font_name": "4x6",
```
ty-porter, line 191: Just realized this will override the master font, so will need to update these where added.

coordinates/w32h32.json.example
```diff
@@ -225,6 +227,14 @@
...
         "x": 64
       }
+    },
+    "postseason": {
+      "matchup_y_gap": 12,
+      "series_x_gap": 32,
+      "wc_x_start": 3,
+      "wc_y_start": 12,
+      "ds_a_y_start": 51,
+      "__comment": "all other coords are based off wild card position"
```
ty-porter, line 237: Open to suggestions on this... Didn't look, but seems like if this is here then postseason brackets render, so may need to add an explicit `enable`/`disable` flag

coordinates/w64h64.json.example
```diff
@@ -105,7 +105,15 @@
...
-    "y": 46
+    "y": 46,
+    "nohit": {
+      "x": 46,
+      "y": 46
+    },
+    "perfect_game": {
+      "x": 46,
+      "y": 46
+    }
```
ty-porter, line 116: I don't believe this is used anymore, but it's in most of the coordinates examples, so I added here.

WardBrian:
I’d be nice to validate against a schema but having knowledge of the no-hitter/perfect game ideas without copying those from wherever they are in the schema. Probably also ignoring keys named “__comment” etc

coordinates/w64h64.json.example
WardBrian: My custom layout uses this feature pretty extensively but not in any of the places the example does

PR #451, opened 2023-04-06
Update readme

tracerrx:
Updated readme for upcoming release

### DIFF
--- a/coordinates/README.md
+++ b/coordinates/README.md
@@ -30,3 +30,4 @@ The software develops and releases features with full support for the default la
 A couple of things are not completely implemented or have some implementation details you should understand.
 
 * `bases` currently requires an even `size` value to be rendered correctly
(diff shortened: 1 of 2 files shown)


coordinates/README.md
```diff
@@ -30,3 +30,4 @@ The software develops and releases features with full support for the default la
 A couple of things are not completely implemented or have some implementation details you should understand.
 
 * `bases` currently requires an even `size` value to be rendered correctly
+* Not all options are enabled on all board sizes by default.  For example pitch count and pitch type are not enabled by default on boards smaller than 64x64.
```
ty-porter, line 33: I would mention how this is disabled since that was cause for confusion:

```
Options are "disabled" by forcing them to render outside the board, by setting X and Y coordinates less than 0 or greater than the height or width of the board.
```

ty-porter (approved):

PR #452, opened 2023-04-06
Fix preferred team 

tracerrx:
A bugfix to allow for scrolling/selection of second/third preferred team even if first is not playing today.

### DIFF
--- a/data/schedule.py
+++ b/data/schedule.py
@@ -126,14 +126,19 @@ class Schedule:
 
     def _game_index_for_preferred_team(self):
         if self.config.preferred_teams:
-            team_name = data.teams.TEAM_FULL[self.config.preferred_teams[0]]
-            team_index = self.current_idx
-            team_idxs = [i for i, game in enumerate(self._games) if team_name in [game["away_name"], game["home_name"]]]
-            if len(team_idxs) > 0:
-                team_index = next((i for i in team_idxs if status.is_live(self._games[i]["status"])), team_idxs[0],)
-            return team_index
-        else:
-            return self.current_idx
+            team_idxs = []
+            for team in self.config.preferred_teams:
+                team_name = data.teams.TEAM_FULL[team]
+                for i, game in enumerate(self._games):
+                    if team_name in [game["away_name"], game["home_name"]]:
+                        if status.is_live(game["status"]):
+                            team_idxs.append(i)
+                            break
+                        elif i == len(self._games) - 1:
+                            team_idxs.append(i)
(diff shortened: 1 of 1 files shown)


WardBrian:
What is the intended behavior here? If your number 1 preferred team isn't playing but a later one is, the first one on that list becomes "sticky" and doesn't rotate?

This is sort of tricky to change because I quite like the way I currently have things. My config is:

Nats as my preferred team
Several other teams I follow listed afterwards

Rotation is set to "only preferred" and enabled only during inning breaks for my preferred team (the Nats).

So, if the Nats are playing, I always see their game. During inning breaks, or if they are not playing, it will rotate through the Red Sox, Mets, etc. (my other preferred teams). 

After this PR, it sounds like rather than rotating through those other teams, it would always prefer to show one of them. Is that correct?

Assuming so, we probably want to split this idea into something like "preferred teams" and "followed teams", with the second referring to teams I want to see in the rotation but never really want to become the "favorite"/sticky game

tracerrx:
Maybe, whats happens now is that if your primary preferred team is off today, none of the other preferred team games show. It only works if your primary preferred team is playing. Set you config to yankees and blue jays, you would expect to see the toronto game but you don't because the yankees game was postponed.

WardBrian:
I think that’s only true if you have one or both of the “team_offday” config settings enabled

WardBrian:
I have it set it to show on league off days only, not team off days. These are pretty rare during the season, mostly things like the All Star Break

tracerrx:
I think what were looking for is a mix.. from your above post I agree it should be: "So, if the Nats are playing, I always see their game. During inning breaks, or if they are not playing, it will rotate through the Red Sox, Mets, etc. (my other preferred teams)." But when there are no games active (i.e. in the morning) it should show weather etc.  Currently it will do that if you have news ticker team offday set to true only, but then it misses your 2nd/3rd team games.

WardBrian:
Yeah, this is why I added the ability for standings to show if no game was currently life (rather than if it was an off day). Something similar for the News/Weather is probably what we want

tracerrx:
So lets close this PR.. and later on well add "display_no_games_live" functionality to both weather and news... I didn't realize that was there for standings...

WardBrian:
If we can make it so weather and news aren’t the select-once-and-stay-there style they currently are, we’re like 70% of the way to letting you pick whatever rotation order you want, which would be sweet. It would be cool to show the standings during the inning breaks, or the weather

PR #453, opened 2023-04-07
Create w192h64.json.example

peteywhit:
Created a working example layout file for 192x64 screen size based on the 128x64 example.

YrVMjSWc

### DIFF
--- a/coordinates/w192h64.json.example
+++ b/coordinates/w192h64.json.example
@@ -1,0 +1,343 @@
+{
+  "defaults": {
+    "font_name": "7x13"
+  },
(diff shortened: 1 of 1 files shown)


WardBrian:
This looks pretty reasonable to me. The periods on the standings overlap the lines between teams, so maybe the font needs to be a bit smaller there?

A few things like the postseason bracket could probably be more centered, it mostly looks the same as the 128x64

ty-porter (approved):
I say let's ship this and deal with minor issues as they arise 👍🏻

PR #454, opened 2023-04-12
Fixed capital N in 4x6 font

tracerrx:
Patched N in 4x6 font.


ty-porter:
Reverted #416 which introduced this issue

PR #455, opened 2023-04-15
Allow news ticker when no games are live

WardBrian:
This allows you to have `no_games_live` set for the news and weather screen, just like the current setting for standings.

If both are set, it will rotate between the two every couple minutes until games become live. 

This is accomplished with a kind of clever system where we can pass in conditions from the outside which control how long each screen is shown for.  This also starts to pave the way for a better overall rotation config. I'd love to eventually allow you to have a rotation which looks like this:

```
offday: [news]                                           # options: news, standings
no_games_live: [news, standings]                         # options: news, standings
gameday: [preferred_teams, news, standings, live_games]  # options: preferred_teams (never rotate during these except inning breaks), news, standings, followed_teams (teams you follow but don't want to become sticky like preferred teams), live_games (all other games)
```

Not sure if I'll have a go at implementing that here or later on.

### DIFF
--- a/renderers/main.py
+++ b/renderers/main.py
@@ -56,7 +56,7 @@ class MainRenderer:
 
         if news and standings:
             while True:
-                self.__draw_news(timer_cond(STANDINGS_NEWS_SWITCH_TIME, refresh=self.data.config.scrolling_speed))
+                self.__draw_news(timer_cond(STANDINGS_NEWS_SWITCH_TIME))
                 self.__draw_standings(timer_cond(STANDINGS_NEWS_SWITCH_TIME))
         elif news:
             self.__draw_news(permanent_cond)
@@ -77,9 +77,7 @@ class MainRenderer:
         while True:
             if not self.data.schedule.games_live():
                 if self.data.config.news_no_games and self.data.config.standings_no_games:
-                    self.__draw_news(
-                        all_of(timer_cond(STANDINGS_NEWS_SWITCH_TIME, refresh=refresh_rate), self.no_games_cond)
-                    )
+                    self.__draw_news(all_of(timer_cond(STANDINGS_NEWS_SWITCH_TIME), self.no_games_cond))
                     self.__draw_standings(all_of(timer_cond(STANDINGS_NEWS_SWITCH_TIME), self.no_games_cond))
                     continue
                 elif self.data.config.news_no_games:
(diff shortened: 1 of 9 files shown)


WardBrian:
I made the somewhat natural extension of allowing news/standings settings for offdays to compose. So now if you set `team_offday` to true for both, it will switch between them rather than picking one

renderers/main.py
```diff
@@ -214,16 +173,127 @@ def __draw_game(self):
...
+    return True
+
+
+def timer_cond(seconds, refresh=1) -> Callable[[], bool]:
+    """Create a condition that is true for the specified number of seconds"""
+    curr = 0
+
+    def cond():
+        nonlocal curr
+        curr += refresh
```
ty-porter, line 286: This doesn't seem accurate. Why not just check current timestamp - started timestamp and compare to target?

renderers/main.py
```diff
@@ -29,104 +33,59 @@ def __init__(self, matrix, data):
...
+                if self.data.config.news_no_games and self.data.config.standings_no_games:
+                    self.__draw_news(
+                        all_of(timer_cond(STANDINGS_NEWS_SWITCH_TIME, refresh=refresh_rate), self.no_games_cond)
+                    )
+                    self.__draw_standings(all_of(timer_cond(STANDINGS_NEWS_SWITCH_TIME), self.no_games_cond))
+                    continue
+                elif self.data.config.news_no_games:
+                    self.__draw_news(self.no_games_cond)
+                elif self.data.config.standings_no_games:
+                    self.__draw_standings(self.no_games_cond)
```
ty-porter, line 88: Do we need to `continue` out of these 2 conditions also?

ty-porter (commented):
In general it seems like it makes sense, I can review more thoroughly as it moves out of WIP

renderers/main.py
WardBrian: Not necessarily- when those functions return, it's because a game is live, so we do want to go directly into drawing it.

renderers/main.py
WardBrian: This is exactly the kind of question I love code review for. The answer to "why not" is "I didn't think of that".

Way better.

WardBrian:
@ty-porter I don't think I'm going to push this any further toward that full refactor for the time being, so I'd consider this ready for the full review.

The only open question is if we want to move the `STANDINGS_NEWS_SWITCH_TIME` constant into a configurable value or not.

WardBrian:
My board crashed today and I suspected this was to blame, but it was actually an unrelated issue. My board tried to display a Wild Card standing at the exact second a refresh happened, and during that exact second the list of divisions hadn't been correctly populated. 

I fixed this by storing it as a local until it is fully populated and then assigning it to the member variable

ty-porter (approved):
This looks good to me on second pass.

PR #456, opened 2023-04-15
Add the ability to truncate team names to a desired length

swemoney:
Currently working on a 64x32 layout that uses the 5x7 font a lot more. Needed a way to specify a specific number of characters to allow for the team name (instead of just going to the short 3 character names). This will allow you to add a `truncate` key to the `teams.name.home` or `teams.name.away` key-paths in your coordinates file to specify a truncate length. If the key isn't found, it falls back to the previous default of 13 characters.

### DIFF
--- a/renderers/games/teams.py
+++ b/renderers/games/teams.py
@@ -107,7 +107,10 @@ def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, 
     font = layout.font("teams.name.{}".format(homeaway))
     team_text = "{:3s}".format(team.abbrev.upper())
     if full_team_names:
-        team_text = "{:13s}".format(team.name)
+        truncate_length = 13
+        if layout.state_is_warmup() == False:
+            truncate_length = layout.coords("teams.name.{}".format(homeaway)).get("truncate", 13)
+        team_text = team.name[0:truncate_length]
     graphics.DrawText(canvas, font["font"], coords["x"], coords["y"], text_color_graphic, team_text)
 
 
--- a/renderers/games/teams.py
+++ b/renderers/games/teams.py
@@ -107,9 +107,7 @@ def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names, 
     font = layout.font("teams.name.{}".format(homeaway))
     team_text = "{:3s}".format(team.abbrev.upper())
     if full_team_names:
-        truncate_length = 13
-        if layout.state_is_warmup() == False:
-            truncate_length = layout.coords("teams.name.{}".format(homeaway)).get("truncate", 13)
+        truncate_length = layout.coords("teams.name.{}".format(homeaway)).get("truncate", 13)
         team_text = team.name[0:truncate_length]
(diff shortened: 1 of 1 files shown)


renderers/games/teams.py
```diff
@@ -107,7 +107,10 @@ def __render_team_text(canvas, layout, colors, team, homeaway, full_team_names,
     font = layout.font("teams.name.{}".format(homeaway))
     team_text = "{:3s}".format(team.abbrev.upper())
     if full_team_names:
-        team_text = "{:13s}".format(team.name)
+        truncate_length = 13
+        if layout.state_is_warmup() == False:
```
legobmw99, line 111: Is there a reason this wouldn't happen during the warmup?

WardBrian:
It seems pretty config specific to not have this take effect during the warmup statuses. Not sure if there is a general way to capture that idea however.

How does this end up looking in practice?

swemoney:
Ya. Maybe I can try and add the warmup key to my teams.name paths and see if I can specify a specific truncate length for that. I only ignored during warmup since the runs/errors/hits leave more room to display a full team name.

swemoney:
I'll need to wait for tomorrow for another warmup game to test it out

WardBrian:
I forgot we had warmup as a state (I only use the no hitter/perfect game states in my config) but it should definitely be possible to use that for this and not have that conditional.

swemoney:
It's been a while since I worked on this codebase and so much has changed. Just needed to brush off some dust to remember how the optional keys worked again. 

Removed the conditional. Now you'll need to add a full `warmup` key in addition to the `truncate` key to get this to work. But it's completely optional and everything should work exactly like it did before if you don't need to worry about truncating.

Example:
```...
    "name": {
      "away": {
        "font_name": "5x7",
        "x": 2,
        "y": 7,
        "truncate": 8,
        "warmup": {
          "font_name": "5x7",
          "truncate": 13,
          "x": 2,
          "y": 7
        }
      },
      "home": {
        "font_name": "5x7",
        "x": 2,
        "y": 15,
        "truncate": 8,
        "warmup": {
          "font_name": "5x7",
          "truncate": 13,
          "x": 2,
          "y": 15
        }
      }
    },
    ...```

WardBrian (approved):
Seems reasonable. Do you think we should add this field (with the default value of 13?) to the example layout files for discoverability?

swemoney:
I'm not sure if it's too niche of a thing to stick it in all of the examples, but maybe we could document it in the coords wiki? Either one could work though.

WardBrian:
Maybe you could contribute your eventual layout which uses it to https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/wiki/Community-Layouts-and-Colors and that would be enough?

swemoney:
That could be enough. If we have other users craving a way to truncate team names in the layouts, we could always do more for discoverability then.

swemoney:
Did these changes get overwritten somewhere? I'm not seeing them in master or dev right now. I know I messed up by branching off of master instead of dev initially. Not sure if that screwed up the merge somewhere.

PR #460, opened 2023-04-16
Workflow to require release branch to merge to master

ty-porter:
Ironically this PR needs to be pointed to master 🙃

This PR will add the workflow to prevent non `dev` branch merges to master. This workflow will be added as a branch protection rule for master, which can be overridden with specific permissions if there's an emergency fix required.

Should cut down on instances of accidental merges to master and will let contributors have better visibility into where they should branch from


WardBrian:
This will be very useful (my bad on the latest example!)

Issue #485, opened 2023-07-08
Option to display team record on game screen


ty-porter:
## This issue is a

- [ ] Bug Report
- [x] Feature Request

From Discord:

```
July 6, 2023 | 8:20 PM]Kevin M: hey, crew! I know I'm in the minority,
but I prefer the team initials display to the full team name.
Since there's a lot of empty space between the name and the score data,
could we add an option to display team record after it? 
BAL (50-35)  8 12 0
NYY (48-39)  0  1  0
```

ty-porter:
Released in v6.4.0

ChickenSupreme:
I just enabled this feature on my 64x32 board and it's great! Unfortunately when both runs AND hits are >9 the record overlaps the runs and when just runs OR hits are >9 there's no empty space between them at all (and once a record includes a 3 digit number that case will overlap as well).

I decided to make a quick edit to only display the record when runs and hits are both <10. It's a pretty simple change for anyone else who wants to do the same:

Open renderers/games/team.py and look for this code (currently lines 64-65):

```
    __render_record_text(canvas, layout, away_colors, away_team, "away", default_colors, away_name_end_pos)
    __render_record_text(canvas, layout, home_colors, home_team, "home", default_colors, home_name_end_pos)
```

Change it to:

```
    if can_use_full_team_names(canvas, True, True, [home_team, away_team]):
        __render_record_text(canvas, layout, away_colors, away_team, "away", default_colors, away_name_end_pos)
        __render_record_text(canvas, layout, home_colors, home_team, "home", default_colors, home_name_end_pos)
```

I have no idea if there's enough demand for this to be added as a configurable option and this particular change is a bit of a hack anyway but I figured I'd leave this note here in case anyone else wants to apply it to their board!

As always thanks for the work you guys do on this. It's a very cool project.

PR #486, opened 2023-07-08
Add option to display team record on game screen

ty-porter:
Addresses #485

Still undergoing testing. This adds the layout keys needed to render the team record on the game screen. By default, the record will render off screen since it takes up quite a bit of space (8 max):

```
(WWW-LL)
12345678
```

Requested behavior can be achieved with a custom coordinates file and the following config setting:

```
 "full_team_names": false
```

Future enhancements may need to specify smaller font sizes to fit in smaller

### DIFF
--- a/coordinates/w128h32.json.example
+++ b/coordinates/w128h32.json.example
@@ -291,6 +291,17 @@
         "y": 10
       }
     },
+    "record": {
+      "_comment": "Offscreen by default.",
+      "away": {
+        "x": -10,
+        "y": -10
+      },
+      "home": {
+        "x": -10,
+        "y": -10
+      }
+    },
     "runs": {
       "runs_hits_errors": {
         "show": false,
--- a/coordinates/w128h32.json.example
+++ b/coordinates/w128h32.json.example
@@ -292,14 +292,16 @@
       }
     },
     "record": {
-      "_comment": "Offscreen by default.",
+      "enabled": false,
       "away": {
-        "x": -10,
-        "y": -10
+        "font_name": "4x6",
+        "x": 19,
+        "y": 7
       },
       "home": {
-        "x": -10,
-        "y": -10
+        "font_name": "4x6",
+        "x": 19,
+        "y": 17
       }
     },
     "runs": {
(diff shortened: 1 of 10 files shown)


WardBrian:
I was working on a version here: https://github.com/WardBrian/mlb-led-scoreboard/tree/feature/pregame-record

as the name suggests I was only having it on the pregame screen, but I stopped really developing after I realized it wouldn’t fit my layout

ty-porter:
Screenshots of all officially supported boards:

Size | Screenshot | Notes
--- | --- | ---
w32h32 | w32h32 | Not intending to provide good support for this size
w64h32 | w64h32 | Will probably overlap when RHE >10, unless H/E disabled like w64h64
w64h64 | w64h64-4x6font | Font 4x6 override
w128h32 | w128h32-4x6font | Font 4x6 override
w128h64 | w128h64
w196h64 | w196h64

WardBrian:
I think any board at least 64-wide would be able to show them during the pregame when there are no scores

ty-porter:
It looks like the main renderer is checking `status.is_pregame(game.status())` for each game and rendering out of the pregame renderer which can either do a warmup state or a start time if not in warmup

It should be fine for the team renderer to make that check as well

WardBrian:
I meant the `data.config.layout` for warmup:
https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/5dcc41f49d2480feba8e1c35f5eea045366b952e/data/config/layout.py#L8

I don't think anyone uses this, but in theory it works like the no-hitter layout where you can do custom coordinates for that state of game but not others. It looks like this is literally done only for the `Warmup` status and not all pregame statuses however

ty-porter:
> It looks like this is literally done only for the Warmup status and not all pregame statuses however

I think this is my hangup here, I'm not convinced adding this for only the `warmup` state will give a consistent experience.

If we need that I think that can be a follow-on?

WardBrian:
As long as there is an "enabled" flag in the coordinates, individual users could enable it in the warmup layout state if they wanted to. I think we should consider adding more layout states like "pregame", but that can definitely be a follow-on

Issue #471, opened 2023-05-11
Add result of previous play to game screen


ty-porter:
## This issue is a

- [ ] Bug Report
- [x] Feature Request

## Additional Information
We display an animation for strikeouts and strikeout looking as well as HR. The `eventType` field also includes info about the previous play

We should be able to add simple cases like single/double/triple/strikeout/walk, etc.

We should avoid play results that require input from the official scorer (such as error) because there is an inherent delay in the scorer's decision

Can grab available event types:
```python
statsapi.meta("eventTypes")
```

PR #491, opened 2023-07-15
Add result of previous play to game screen

jimcreel:
Initial solution for Issue #471: Add result of previous play to game screen

This adds results for hits, walks, and strikeouts to the game screen, using the logic previously in place to display strikeouts. 

Displays a flashing long or short description (based on settings in coords for pitch descriptions) in the batter's text area

The __render_play_result method replaces the __render_strikeout method to handle all play results that will trigger an animation as defined in data/plays.py

The spacing could probably look better in terms of how it displays on the screen, but it is readable and a start!

MLB-LED-SCOREBOARD


data/plays.py
```diff
@@ -0,0 +1,36 @@
+PLAY_RESULT_UPDATES = ["single", "double", "triple", "walk", "intent_walk", "home_run", "strikeout", "strikeout_looking"]
```
WardBrian, line 1: There are other play results than just these - not that we'd want to show them necessarily, but code which uses this file should keep in mind that indexing into this or the dictionary can fail

data/plays.py
```diff
@@ -0,0 +1,36 @@
...
+
+PLAY_RESULTS ={
```
WardBrian, line 3: Can we split this into two dictionaries rather than a nested dictionary?

renderers/main.py
```diff
@@ -153,7 +153,7 @@ def __draw_game(self):
                 self.data.scrolling_finished = True
 
         else:  # draw a live game
-            if scoreboard.homerun() or scoreboard.strikeout():
+            if scoreboard.homerun() or scoreboard.strikeout() or scoreboard.hit() or scoreboard.walk():
```
WardBrian, line 156: Does hit and walk lead to flashing text like the strikeout does?

colors/scoreboard.json.example
```diff
@@ -124,7 +124,7 @@
       "g": 255,
       "b": 255
     },
-    "strikeout": {
+    "play_result": {
```
WardBrian, line 127: I think I'd want Ks and hits to be two different colors, if we could allow them to be split

WardBrian (commented):
Some general thoughts - not all necessarily requirements for merging this IMO, just to keep in mind

data/plays.py
jimcreel: Agreed. I think the issue might be the naming of the list here. These are meant to be specific types of play results that will trigger the display change. Would naming this list appropriately address the concern, or do we want to move the logic to the method, only sending the sorts of play results that we want to trigger?

renderers/main.py
jimcreel: Yes, they do. During the first draft I didn’t intend it, but I kind of liked it during testing and it grew on me, especially for home runs. I would be ok with it either way, but will defer to y’all.

colors/scoreboard.json.example
jimcreel: Sounds good to me. Red for K’s and yellow for everything else? Or did you have something else in mind?

data/plays.py
ty-porter: I think the top array is redundant, we can get the same with either of:

```python
# dict_keys object, iterable but doesn't have subscription
PLAY_RESULTS.keys()

# list
list(PLAY_RESULTS.keys())
```

Only downside to the top one is you can't do the following, but this PR doesn't use it

```python
PLAY_RESULTS.keys()[0]
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# TypeError: 'dict_keys' object is not subscriptable
```

data/plays.py
ty-porter: Would prefer these to be constants, they are referred all throughout the code as magic strings. If we're going to index into a dict this should be mapped here:

```python
SINGLE   = "single"
DOUBLE   = "double"
TRIPLE   = "triple"
HOME_RUN = "home_run"

WALK             = "walk"
INTENTIONAL_WALK = "intent_walk"

STRIKEOUT         = "strikeout"
STRIKEOUT_LOOKING = "strikeout_looking"

HITS = [
  SINGLE,
  DOUBLE,
  TRIPLE
]

WALKS = [
  WALK,
  INTENTIONAL_WALK
]

STRIKEOUTS = [
  STRIKEOUT,
  STRIKEOUT_LOOKING
]
```

data/scoreboard/__init__.py
```diff
@@ -41,6 +41,12 @@ def strikeout(self):
 
     def strikeout_looking(self):
         return self.play_result == "strikeout_looking"
+    
+    def hit(self):
+        return self.play_result == "single" or self.play_result == "double" or self.play_result == "triple"
```
ty-porter, line 46:
```suggestion
        return self.play_result in ["single", "double", "triple"]
```

The other comment I made constantizes the plays, I'd really like to do just this, depending on how it needs to get imported

```python
        return self.play_result in data.plays.HITS
```

renderers/games/game.py
```diff
@@ -46,27 +47,26 @@ def render_live_game(canvas, layout: Layout, colors: Color, scoreboard: Scoreboa
...
-    coords = layout.coords("atbat.strikeout")
-    color = colors.graphics_color("atbat.strikeout")
-    font = layout.font("atbat.strikeout")
-    text = "ꓘ" if looking else "K"
+def __render_play_result(canvas, layout, colors, play_result):
+    
+    coords = layout.coords("atbat.play_result")
+    color = colors.graphics_color("atbat.play_result")
+    font = layout.font("atbat.play_result")
+    text = PLAY_RESULTS[play_result][coords["desc_length"]]
```
ty-porter, line 67: This is a brittle lookup, it will fail if `play_result` or the result of `coords["desc_length"]` can't be found

Lots of ways to handle it, but easiest is `try`/`catch` on `KeyError`

```suggestion
    try:
      text = PLAY_RESULTS[play_result][coords["desc_length"]]
    except KeyError:
      return # There's no text or coordinates to render
```

coordinates/w64h32.json.example
```diff
@@ -139,7 +139,18 @@
...
     },
     "loop": 36,
-    "strikeout": {
+    "play_result": {
+      "x": 15,
+      "y": 29,
+      "font_name": "5x7",
+      "desc_length": "Short"
+    },
+    "hit": {
```
ty-porter, line 148: These two look like they shouldn't be here?

data/plays.py
ty-porter: Also if we constantize the plays we should do the same for this dict:

```python
PLAY_RESULTS = {
  SINGLE: {
    "Short": "1B", 
    "Long": "Single"
  },
  DOUBLE: {
    "Short": "2B", 
    "Long": "Double"
  },
  TRIPLE: {
    "Short": "3B", 
    "Long": "Triple"
  }
  # . . .etc.
} 
```

ty-porter: And last nitpick:

I prefer lowercase for key names:

``` python
# Current
{ "Short": "1B", "Long": "Single" },

# Preferred
{ "short": "1B", "long": "Single" },
```

data/plays.py
jimcreel: Got it! It's done now.

data/plays.py
jimcreel: Also done. The key was capitalized to match the style of the coords json which uses upper case (to match the style of the pitches coords), but I changed the coords file (only for plays) rather than altering the string to match.

coordinates/w64h32.json.example
jimcreel: Yes, that's a vestige from before I decided to lump them all under "play_result". It's fixed in the latest commit

ty-porter (approved):

colors/scoreboard.json.example
WardBrian: Yeah that makes sense as a default. The important thing is letting them be customized

renderers/main.py
WardBrian: With home runs already having the bases circling animation it may be a bit much, but I’m happy to give it a try. 

It’s also totally reasonable for us to circle back and add more knobs for customization of this if we want to later

colors/scoreboard.json.example
jimcreel: Sounds good, it's implemented now

PR #500, opened 2023-10-02
Fix: postseason bracket rotation

WardBrian:
The logic here was a little broken which caused a crash when the bracket was displayed


WardBrian:
This will start crashing boards in the morning unfortunately @ty-porter

ty-porter (approved):

PR #540, opened 2024-05-20
Update custom driver implementation with encapsulated RGBMatrixDriver

ty-porter:
Replaces this temporary spike branch (https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/tree/fps-counter)

New driver implementation includes display overlay or log format FPS counter


WardBrian (commented):
Seems like a nice clean up -- I think this needs a new addition to install.sh and requirements.txt, right?

ty-porter:
> Seems like a nice clean up -- I think this needs a new addition to install.sh and requirements.txt, right?

I think the installer is fine if you are referring to the `--emulator-only` flag, it's a convenience flag for skipping installation with `sudo` and not installing the hardware matrix driver

Requirements updated, replaced emulator with driver and appropriate stable version:

[requirements.txt](https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/pull/540/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552)

ty-porter:
Going to close this, the driver lib I wrote works great with the emulator but the extra features break the hardware implementation. That's more work than I want to resolve at the moment

PR #548, opened 2024-07-30
Additional play display compatibility

Xatunix:
This pull request makes changes in data/plays.py and data/game.py to add compatibility with numerous other common plays in the scoreboard. This could be combined with changes in renderers/game/game.py to display many more play results on scoreboards (as I've done with the rest of the code on my personal repository).

Specifically:
- Added "RBI" tag for run scoring plays to differentiate them for others
- Added "Home Run" play for display of all types of Home Runs
- Added tag for type of out made (fly, pop, ground) to allow for display
- Added other plays such as "Stolen Base"

All original plays and abbreviations should have been maintained and not break any existing functionality.

(I don't know what I'm doing, so please provide feedback if going about this incorrectly)


data/game.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, game_id, date, broadcasts, series_status, delay_in_10s_of_sec
         self._broadcasts = broadcasts
         self._series_status = series_status
         self._status = {}
+        print(self.game_id)
```
WardBrian, line 48:
```suggestion
```

data/game.py
```diff
@@ -314,6 +315,105 @@ def current_play_result(self):
...
+        elif result == "field_out" and (
+            "lines"
+            in self._current_data["liveData"]["plays"].get("currentPlay", {}).get("result", {}).get("description", "")
+        ):
+            result += "_line"
+        elif result == "field_out" and (
+            "pop"
+            in self._current_data["liveData"]["plays"].get("currentPlay", {}).get("result", {}).get("description", "")
+        ):
+            result += "_pop"
```
WardBrian, line 337: I think it would make sense to turn this into a helper function called like `out_type`

data/game.py
```diff
@@ -314,6 +315,105 @@ def current_play_result(self):
...
+        elif result == "walk" and (
+            "scores"
+            in self._current_data["liveData"]["plays"].get("currentPlay", {}).get("result", {}).get("description", "")
+        ):
+            result += "_rbi"
+        elif result == "hit_by_pitch" and (
+            "scores"
+            in self._current_data["liveData"]["plays"].get("currentPlay", {}).get("result", {}).get("description", "")
+        ):
+            result += "_rbi"
```
WardBrian, line 400: Similarly, I think making a helper function called "did_score" or something could simplify this

WardBrian (changes requested):
Thanks for contributing! Here are some of my thoughts on this as it stands

Xatunix:
Thanks for the detailed thoughts (some of which made me facepalm)! I'll revise accordingly.

PR #556, opened 2024-09-03
Allow better name customization

WardBrian:
This changes the logic which displays names or abbreviations to always reference teams.py, which is built on top of team ID as the "base" lookup value.

As a result, it is easier to use custom abbreviations or names for teams.

See discussion: https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/discussions/554


data/teams.py
```diff
@@ -1,72 +1,109 @@
...
-    "Rockies": "Colorado Rockies",
-    "Tigers": "Detroit Tigers",
-    "Astros": "Houston Astros",
-    "Royals": "Kansas City Royals",
-    "Dodgers": "Los Angeles Dodgers",
-    "Nationals": "Washington Nationals",
-    "Mets": "New York Mets",
+# Can be customized, but will require changing in colors/teams.json as well
+TEAM_ID_ABBR = {
+    133: "OAK",
```
ty-porter, line 10:
```suggestion
    133: "ATH",
```

😢

ty-porter (approved):

PR #561, opened 2024-11-04
Allow selection of driver version and set 14ab2ff as current default

ty-porter:


WardBrian:
Shouldn’t this be a command line option to the install script, not a config entry?

ty-porter:
Yeah we can default it in the installer. I think it's a little easier from end user's perspective to be able to put it into a config, but most people won't be tweaking that often, so this way is probably better long-term.

WardBrian (approved):

PR #569, opened 2025-02-13
Release v8.2.0

ty-porter:


WardBrian:
Do we want to bump the version of the matrix driver in this release? And maybe revert to Tom-thumb 4x6 font (I’ve always preferred it personally)

ty-porter:
> Do we want to bump the version of the matrix driver in this release? And maybe revert to Tom-thumb 4x6 font (I’ve always preferred it personally)

We can give it a stab, I didn't realize how out of date the commit we currently use is. We're gonna need someone to test this I think

https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/67ff79d96044d0212bc6db15fe2115efcfb8bcfc/install.sh#L8

https://github.com/hzeller/rpi-rgb-led-matrix/compare/14ab2ff...master

PR #571, opened 2025-02-18
Move rpi-rgb-led-matrix version to master

ty-porter:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -83,6 +83,7 @@ if [ "$SKIP_PYTHON" = false ]; then
         python3-pillow \
         python3-tk \
         python3-venv \
+        cython3 \
         libxml2-dev \
         libxslt-dev \
         libsdl2-mixer-2.0-0 \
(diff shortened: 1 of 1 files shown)


WardBrian:
When testing locally I needed to also add pip install cython and a CYTHON= to my make command, but otherwise I didn’t need anything else

ty-porter:
Looks like it

https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/bindings/python/README.md#python-3

```sh
sudo apt-get update && sudo apt-get install python3-dev cython3 -y
make build-python 
sudo make install-python 
```

WardBrian:
We can `$PYTHON -m pip install cython`
`make build-python PYTHON=$PYTHON CYTHON=cython` (cython from pip is version 3 but doesn't call itself cython3)

if we don't want to globally install it. Not sure it matters either way

install.sh
```diff
@@ -144,7 +145,7 @@ if [ "$SKIP_MATRIX" = false ]; then
     # Checkout the branch or commit specified for rpi-rgb-led-matrix
     git checkout $DRIVER_SHA
     git pull
-    make build-python PYTHON="$PYTHON"
+    make build-python PYTHON="$PYTHON" CYTHON=cython
```
WardBrian, line 148: on my system, the `cython3` apt package installs under `cython3`, not `cython`. Pip, on the other hand, installs under `cython`

ty-porter: Cool, was shooting in the dark til I could find a machine to test on but I got this working with `cython3`

PR #576, opened 2025-02-24
Add tests for game, standings, and schedule objects against API

WardBrian:
This adds a few basic tests. 

They aren't the most elegant thing in the world since we really want to be testing both our own code and the underlying MLB api, but hopefully they will help us catch regressions

### DIFF
--- a/renderers/main.py
+++ b/renderers/main.py
@@ -22,7 +22,7 @@ class MainRenderer:
     def __init__(self, matrix, data):
         self.matrix = matrix
         self.data: Data = data
-        self.is_playoffs = self.data.schedule.date > self.data.headlines.important_dates.playoffs_start_date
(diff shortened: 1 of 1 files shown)


ty-porter (approved):
This is as good a start as any

Have had success with VCR (Ruby library) in the past, seems there is a Python version also
It's nice because it makes the tests are deterministic, handles the response storage automatically, etc.

https://github.com/kevin1024/vcrpy

Not pushing for it in this PR as it comes with the downsides of mocking, but something to think about.
We would also have to consider how we capture and test against live games

WardBrian:
VCRpy looks interesting. I think there is value in actually testing against the APIs in case the schemas change out from under us between seasons, but it's probably worth thinking about both ways

PR #578, opened 2025-02-27
Special uniform color support

### LINKED ISSUE #399: Introduce City Connect colors
## This issue is a

- [ ] Bug Report
- [x] Feature Request

Introduce behavior to use applicable City Connect colors on Tuesdays for home games.

WardBrian:
Closes #399 (city connects). I tried to write this in a general enough way that we could extend it to e.g. throwback or other special uniforms if we want to

Depends on https://github.com/toddrob99/MLB-StatsAPI/pull/154

Draft status until the above is released.

Here's what the [double city connect game](https://www.foxsports.com/stories/mlb/nationals-orioles-will-be-first-teams-to-wear-city-connect-jerseys-against-each-other) looks like:

preview image


renderers/games/teams.py
```diff
@@ -39,17 +18,17 @@ def render_team_banner(
     accent_coords["home"] = layout.coords("teams.accent.home")
 
     for team in ["away", "home"]:
+        color = home_colors['home'] if team == "home" else away_colors['home']
```
ty-porter, line 21: I know we have to match the existing configs for the time being but this was pretty confusing at first glance why `home` is always the key

data/scoreboard/team.py
```diff
@@ -1,8 +1,32 @@
...
+        self.special_uniform = special_uniform
+
+
+    def lookup_color(self, team_colors):
+        """
+        Given a team color object, this returns the colors for the team.
+
+        The result is guaranteed to have the 'home', 'accent', and 'text' keys.
+        Any missing data is filled in with the default colors.
+        If the team has a special uniform, the colors for that uniform are substuted
```
ty-porter, line 21: nitpick

```suggestion
        If the team has a special uniform, the colors for that uniform are substituted
```

data/scoreboard/team.py
```diff
@@ -1,8 +1,32 @@
...
+        """
+        default_colors = team_colors.color("default")
```
ty-porter, line 23: Does this also need to be behind a config?

renderers/games/teams.py
WardBrian: yeah it’s a wart. It should probably be “primary”, “text”, “highlight” as the names

data/scoreboard/team.py
WardBrian: Not sure what you mean - there is a “default” section of the config that can be changed already

ty-porter: I meant config to disable special uniform colors altogether

WardBrian: Oh, I don’t think that’s necessary, because you could just delete those sub-keys or set them equal to the normal team colors and that would effectively disable it as well

ty-porter (approved):

PR #580, opened 2025-03-01
Add BDF font metadata parsing

ty-porter:
Added all the fonts
* `4x6-legacy` comes from pre v6.0.0 prior to #428 
* `4x6` is reverted (again) back to v6.0.0 - v8.1.0 version
* Fonts now use `bdfparser` to parse their actual metadata and don't rely on filenames
* Tests for the above.


WardBrian (commented):
Looks good to me (though I will continue using the OG font)

PR #588, opened 2025-03-08
Add logfiles and set exit conditions in installer

### LINKED ISSUE #586: Installer errors do not halt installation and can break silently
We have various reports of pieces of the installer failing to work correctly but do not halt execution.

This usually causes us to have to diagnose the root cause after the fact and without complete logs from the installer.

We should set pipefails that halt execution so it is obvious something is wrong.

ty-porter:
closes #586

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,16 +1,22 @@
 #!/bin/bash
 
+# Redirect output to a logfile
+exec > >(tee -a logs/mlbled.log) 2>&1
+
 SKIP_PYTHON=false
 SKIP_CONFIG=false
 SKIP_MATRIX=false
 NO_SUDO=false
 SKIP_VENV=false
 DRIVER_SHA=master
+FORCE=false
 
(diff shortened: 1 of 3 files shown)


logs/.keep
WardBrian, line 1: We should also .gitignore everything else in this folder I think

install.sh
```diff
@@ -1,16 +1,22 @@
 #!/bin/bash
 
+# Redirect output to a logfile
+exec > >(tee -a logs/mlbled.log) 2>&1
```
WardBrian, line 4: IMO it makes sense to have a separate install.log from the runtime log, but I think it’s fine if they are combined

ty-porter: No strong feelings either way, was hoping it makes it easier for users to send us relevant logs since they will all be in one spot

WardBrian: Fair enough

PR #593, opened 2025-03-11
Allow installer to proceed if submodules/matrix already exists

ty-porter:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -154,9 +154,15 @@ fi
 if [ "$SKIP_MATRIX" = false ]; then
     echo "Running rgbmatrix installation..."
     sudo apt-get install -y make gcc g++
-    mkdir submodules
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -154,9 +154,15 @@ fi
...
+
+    if [ -d matrix ]; then
+        echo "'matrix' directory already exists. Assuming rpi-rgb-led-matrix is already installed."
+    else
+        git clone https://github.com/hzeller/rpi-rgb-led-matrix.git matrix
+    fi
+
     cd matrix
     # Checkout the branch or commit specified for rpi-rgb-led-matrix
     git checkout $DRIVER_SHA
```
WardBrian, line 168: for this to work when the directory already existed, we need a `git fetch`

Issue #363, opened 2022-04-18
“Postponed” status not centered on w128, maybe more


twfarley:

tracerrx:
On 128x32 or 128x64?  Seems to be ok on 128x64.

ty-porter:
If someone would like to contribute a fix

https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/dd3de1c1198d864c86e4faa93f8249d567553fb3/coordinates/w128h64.json.example#L231

This value is not 1/2 of 128 and the text is off-center as a result.

PR #599, opened 2025-03-28
Issue #363: status text centered on w128

sjmy:
This is my first real pull request so please let me know if I've done this correctly!

In w128h64.json.example, changed the status text x value from 56 to 64 so it's centered.

Before:
Issue_363_before

After:
Issue_363_after


.github/PULL_REQUEST_TEMPLATE.md
sjmy, line 1: Probably should not have included this file

WardBrian (approved):
Thanks!

PR #612, opened 2025-04-26
New City Connects for 2025

WardBrian:
Eight teams got new city connects, the last of which leaked a couple weeks ago


WardBrian:
I wrote a small program to help editing these, it displays all the teams colors:

image 

[code](https://gist.github.com/WardBrian/bc3e596fa359cb2f627965d9e2371f48)

We could consider making this a test with the raw emulator to track over time. I'd probably need some help from @ty-porter

ty-porter (approved):
This really boils down to a config change, I am hesitant to write tests against config

The raw adapter would be great if we wanted to basically end-to-end test our rendering code, if that's what you're thinking

PR #627, opened 2025-06-26
renames 10s game_delay_multiplier

tracerrx:
Rename preferred_game_update_delay_in_10s_of_seconds to game_delay_multiplier throughout in order to add sanity to previous PR which moves api_refresh_rate top config.


WardBrian:
I realize I'm responsible for the original sin here, but I don't love `game_delay_multiplier` either. I think it's important to keep in the fact that this only really affects your preferred game. `multiplier` is fine, but "of what" seems like a valid question a user might ask

`preferred_game_refresh_lag`?

ty-porter (approved):

PR #629, opened 2025-06-27
Added ability to set headers on statsapi.game() requests

tracerrx:
Added data.headers with a constant API HEADERS.  Updated the statsapi.game() requests in 

```
data/dates.py
data/game.py
data/standings.py
data/teams.py
data/unifiorms.py
```

Need Brian Ward to take a look at the main game.py requests and replace his `testing_params`

I'm honestly not sure how to test that were actually sending the correct headers

### DIFF
--- a/data/dates.py
+++ b/data/dates.py
@@ -7,11 +7,11 @@ import data.headers
 class Dates:
     def __init__(self, year: int):
         try:
-            data_d = statsapi.get("season", {"sportId": 1, "seasonId": year}, data.headers.API_HEADERS)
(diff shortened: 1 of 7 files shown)


WardBrian:
For the testing headers, you should be able to leave them with just `|` after the params, like 
```python
live_data = statsapi.get("game", {"gamePk": self.game_id, "fields": API_FIELDS} | testing_params , data.headers.API_HEADERS)
```

ty-porter (approved):

Issue #638, opened 2025-07-16
Fix All Star Team Names


tracerrx:
In the discord chat below, during the all star game, the team names overwrite the RHE area because no SHORT NAMES exist.  Either create them, or enable team name scrolling.

https://discord.com/channels/1091751372069806183/1091763502378651768/1394880787261685871

PR #640, opened 2025-07-17
Add All-Star abbreviation and method to lookup teams

ty-porter:


data/teams.py
```diff
@@ -1,112 +1,47 @@
-# the following were created like so
-# import statsapi
-# _teams = statsapi.get('teams', {'sportIds':1})['teams']
-# TEAM_ID_ABBR = {t["id"]: t["abbreviation"] for t in _teams}
-# TEAM_ID_NAME = {t["id"]: t["teamName"] for t in _teams}
-# TEAM_NAME_ID = {t["teamName"]: t["id"] for t in _teams}
-
-# Can be customized, but will require changing in colors/teams.json as well
```
WardBrian, line 8: Should we still mention customization? I know at least one user who uses names like “Cubbies” and “Yanks” instead

ty-porter:
Closes #638

PR #658, opened 2025-11-03
Release v9.0.0

ty-porter:
### Required
- [x] Config migration feature
- [ ] Happy path testing for the config migration feature
- [ ] Rework `validate_config.py`:
  - Should be read-only (changes must come from a migration)
  - Should correctly associate and validate custom configs to its associated schema (ex. `config.json` and `config.alt.json` belong to the same config family referencing `config.schema.json`)
  
### Nice to Have
- [ ] Refactor -- centralize config path strings to a helper that associates configs in the same config family (avoid brittle string interpolation all over the codebase)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -35,6 +35,35 @@ USAGE
     exit 1
 }
 
+handle_error() {
+    local exit_code="$?"
+    local line_number="$LINENO"
+
+    # Red
+    printf "\e[31m"                                                                            >&2
+    echo                                                                                       >&2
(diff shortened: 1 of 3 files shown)


WardBrian:
I can give the config changes and migration code a once over soon if you’re hoping to do this release during the offseason still

ty-porter:
@WardBrian Sounds good. The migration code should be in a decent place. I will update the PR description for what else I'm planning before I consider this ready to release

PR #659, opened 2025-11-17
Add a SQLite framework for settings

ty-porter:


ty-porter:
Abandoning, not much value provided. Requires a ton of work to either build on a UI layer to make this work OR synchronize raw JSON to DB on bootup

PR #665, opened 2026-01-10
Release v8.5.7

ty-porter:


WardBrian:
IMO we can ship this, only question is if we want it to be 8.6 instead

PR #666, opened 2026-02-22
Put render thread in charge of game rotation, allow for standings and weather to be included

WardBrian:
This smooths out a lot of our rendering logic to look much more homogeneous. 

- The main thread no longer decides when the games rotate. Instead it just stores/keeps up to date the current game and the next-to-render game
- The render thread can then decide to move whenever it wants. 
   - This allows us to have the game render code use the same condition system as weather/news, e.g. the games rotate whenever scrolling is finished, a minimum amount of time has passed, and game-specific conditions are met (e.g. checking for liveness if it's the preferred game)
- These make it very natural to add the standings and news headlines as options in the rotation -- once the games have looped around, we can hand off to those two, and then resume the games where we left off

I am pretty sure it re-creates all current configurable properties. There might be a few small behavior changes (like, when the preferred game becomes live after an inning change, you will probably see one extra game rotate through compared to before), but the overall flow of games on the screen seems to be preserved well.

I intend to use this on my personal board throughout ST to test and tweak.


WardBrian:
I think we should probably just merge these changes as part of #674, but I'm leaving this as a breadcrumb of how the code got to where it is in that PR

PR #674, opened 2026-03-07
Update config to allow more complicated rotation rules

### LINKED ISSUE #667: Brainstorming more generic rotation configs
Here's an idea I had while working on #666. 

The config for rotation becomes list of dicts.
Items with `"type": "game"` define rules for deciding which games are considered. The highest non-empty priority is used to fill a list of games.
If there are no non-empty priority levels, priority is `0`.

Any other entries in the rotation list define not a `priority` key, but a `with_priority`. These will be shown as part of the rotation of matching priority levels, but do not introduce new levels.

For example, this following is more or less equivalent to my current default settings, which shows Nationals games and doesn't rotate during them, except for between innings.
All live games rotate when the Nationals aren't live. Standings can appear in live game rotations, and standings and news rotate when nothing is live

```json5
[
  { "type": "game", "priority": 1, "live": true },

  {
    "type": "game",
    "priority": {
      "live": 9, // high enough to basically sticky
      "during_breaks": 1
    },
    "team": "Nationals" // could also be list
  },
  {
    "type": "standings",
    "with_priority": [0, 1] // means it can be shown alongside priority 0 OR 1 items
  },
  {
    "type": "news",
    "with_priority": 0 // means it can be shown alongside priority 1 or 2 items
  }
]
```

The real power is the way this system can define combinations of things we currently can't, like what if I want
Nationals and Dodgers games to rotate with each other, but exclude all other live games unless those
two are off? Easy in this system, impossible now.

### LINKED ISSUE #669: Add Config Options Similar to no_live_games for no_live_or_pregame_games
I'd like to be able to wake up in the morning, see the game scheduled for that day with expected pitchers, etc. and then be able to switch to the "news" and "standings" page once that game is done. Similar to the settings that only show news and standings when there's no live games, I'd like to be able to set the scoreboard to only show news and standings when there's no live _or pre-game_ games. 

I've already implemented this with a config option and am creating this issue so I can make the PR. Will put that up after "E2E testing" with the Tigers spring training game today.

WardBrian:
Would close #667 (see that issue for an overview)
Built on #666 

This is a big one, certainly 9.0 worthy.

This is a new way of controlling what displays on the board based on a `priority` system.

When the board powers on, it is in priority level 0. Two things can change the priority:
1. a `time` rule is matched by the current time. This has no inherent display itself, but itlets you, e.g., show the news screen for an hour or two in the morning
2. a `game` rule is matched. This displays any games matched by this rule, and sets the priority accordingly. 

Once the priority is set, you can _add_ additional games, the news, or standings, which will then all rotate together. 

## Example one: The current default config

The current 8.x config.json is equivalent to
```json
		"screens": [
			{
				"kind": "game",
				"priority": 2,
				"required_status": "live",
				"teams": ["Cubs"]
			},
			{
				"kind": "game",
				"priority": 1
			},
			{
				"kind": "standings",
				"seconds": 60,
				"with_priority": 0
			},
			{
				"kind": "news",
				"seconds": 60,
				"with_priority": 0
			}
		]
```

## Example two: Showing news in the morning, any game later

This also mixes standings in with the games rotations, and shows standings and news if no game is available that day. 

```json
            {"kind":"time", "priority": 2, "start_time": "07:00", "end_time": "09:05"},
            {"kind": "news", "seconds": 20, "with_priority": 2},
            {
                "kind": "game",
                "priority": 1
            },
            {
                "kind": "standings",
                "seconds": 20,
                "with_priority": [0, 1]
            },
            {
                "kind": "news",
                "seconds": 60,
                "with_priority": 0
            }
```

## Example 3: my actual config for testing

This:
1. shows news in the morning
2. treats the US like my top team for WBC purposes
3. Whenever they aren't live, prioritize the nationals
4. If neither are live, show me any live games PLUS the pre/postgame for those two teams only + news + standings
5. if no game is live, show me the pre/post for any game today + news + standings
6. if no game is today, show me just the news and standings

```json

        "screens": [
            {"kind":"time", "priority": 5, "start_time": "07:00", "end_time": "09:05"},
            {"kind": "news", "seconds": 20, "with_priority": 5},
            {
                "kind": "game",
                "priority": 4,
                "required_status": "live_in_inning",
                "teams": ["United States"]
            },
            {
                "kind": "game",
                "priority": 3,
                "required_status": "live_in_inning",
                "teams": ["Nationals"]
            },
            { "kind": "game", "priority": 2, "required_status": "live" },
            {
                "kind": "secondary_game",
                "with_priority": 2,
                "teams": ["United States", "Nationals"]
            },
            { "kind": "game", "priority": 1 },
            {
                "kind": "standings",
                "seconds": 60,
                "with_priority": 0
            },
            {
                "kind": "standings",
                "seconds": 20,
                "with_priority": [1, 2]
            },
            {
                "kind": "news",
                "seconds": 20,
                "with_priority": 2
            },
            {
                "kind": "news",
                "seconds": 60,
                "with_priority": [0, 1]
            }
        ]
```


data/config/__init__.py
```diff
@@ -291,6 +288,47 @@ def __get_layout(self, width, height):
...
+        else:
+            debug.warning(
+                "Invalid screen rule in config, unknown type '{}'. Skipping. Rule: {}".format(
+                    rule_json.get("kind"), rule_json
+                )
+            )
+
+    if not any(screen[0] for screen in screen_rules.values()):
+        # prevents nothing showing for an empty config
+        screen_rules["news"][0] = 60
```
ty-porter: I think this is an error condition that should fail

data/schedule.py
```diff
@@ -1,26 +1,33 @@
...
         self.config = config
         self.date = self.config.parse_today()
         self.starttime = time.time()
         self.current_idx = 0
-        # all games for the day
-        self.__all_games = []
-        # the actual (filtered) schedule
-        self._games = []
+
+        delay_required = ceil(self.config.sync_delay_seconds / GAMES_REFRESH_RATE)
```
ty-porter, line 25: Do we need a delay on the schedule updates? If so, no issues

Seems like the game will have to fill its delay buffer once its rotated to anyway, so there's not much gain here

ty-porter (approved):
Minor comments, this looks good

data/schedule.py
WardBrian: The main reason for the delay here is that, because the schedule is now used for rotating off a game, it’s very easy to get “spoiled” by the `live_in_inning` status rotating away “too early”. The game being displayed might say there are only two outs in the inning, but if the schedule wasn’t delayed it would see that the inning is over and allow you to rotate on

data/config/__init__.py
WardBrian: I was inspired by the old code which would fall back to showing the news if something went wrong in the rest of the render selection, but a hard error would be alright by me

data/schedule.py
ty-porter: Makes sense, thanks.

data/config/__init__.py
ty-porter: Either way works honestly.

I thought an error might be better, otherwise we end up sort of in the same boat we're in now. I can see people asking "why does my board only show news?" repeatedly in Discord

At least if it errors, at worst we might see it in the error log that gets posted

data/utils/double_buffer.py
```diff
@@ -0,0 +1,41 @@
...
+            return self.items[0]
+        else:
+            return self.items[1]
+
+    def consumer_advance(self):
+        if self._active == "current":
+            debug.log("Render thread: requesting main thread to read 'next' game")
+        self._active = "next"
+
+    def producer_tick(self, getter: Callable[[Any], Any]):
```
ty-porter, line 31: noodled on these tick methods overnight. I'm convinced this will work so I don't want this to derail merging it.

ty-porter: I think we can simplify this. `consumer_tick()` / `_producer_acknowledged` seems unnecessary overall

```python
def producer_tick(self, getter: Callable[[Any], Any]):
  if self._active == "current" && (self.items[0] == self.items[1]):
    self.items = (self.items[0], getter(self.items[1])

  if self._active == "next":
    self.items = (self.items[1], self.items[1])
    self._active = "current"
```

`consumer_tick()` goes away -- the producer transparently swaps data in as needed, consumers always read from `active()`, and `advance()` signals that the consumer is ready to advance

Although... Now producers _and_ consumers can write `_active`, so there's possibly a race condition

WardBrian: Yeah, my theory was that by writing it this way, the render thread is the only one that ever writes to `_active`, so it can do it at exactly the places where it knows it isn't accessing `active()`

But I think you're right, this can be simpler. Your suggestion works, but I'm also observing that we call `active()` and `advance()` exactly once per loop, so maybe the right thing to do is fuse them?

WardBrian: Yeah, the specific nature of our loops means we can really just have two methods, `next` called on render and `producer_tick` called on main.

ty-porter: Nice, this looks great. Thanks for the change.

WardBrian: Thank you for the nudge!

PR #664, opened 2026-01-10
Add a migration framework for configs

ty-porter:
Slicing this PR out into its own feature branch for later use.

See original v9.0.0 PR for details
https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/pull/658

PR #677, opened 2026-03-09
Add configuration options for leagues

ty-porter:
This allows us to turn off WBC functionality via config. Disabled by default

Sets us up to support all the sport IDs that StatsAPI supports, notably 2028 summer Olympics

I also extracted a time format helper from #664 primarily to help with emulation crashes on Windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,7 @@ A default `config.example.json` file is included for reference. Copy this file t
 "preferred":                              Options for team and division preference
   "teams"                         Array   An array of preferred teams. The first team in the list will be used as your 'favorite' team. Example: ["Cubs", "Brewers"]
   "divisions"                     Array   An array of preferred divisions that will be rotated through in the order they are entered. Example: ["NL Central", "AL Central"]
+  "leagues"                       Array   An array of preferred leagues. Teams from any league can be included in preferred_teams. Example: ["MLB", "WBC", "Olympics"]. Defaults to ["MLB"]
 
 "news_ticker":                            Options for displaying a nice clock/weather/news ticker screen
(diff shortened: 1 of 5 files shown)


WardBrian:
I’ll have to noodle on this. For one, my refactor in #674 removes the “preferred” key entirely. For another, adding more support for extra leagues is a nontrivial amount of work, and I’m not sure we would ever add anything besides MLB and “International Baseball” (e.g. college)

ty-porter:
This doesn't seem completely incompatible, all it really does is grab more games on the schedule. Where that lives in the config could be anywhere

I think the bigger issue is #674 is enforcing a teams enum which won't work for the national teams. The answer is probably that #674 doesn't need to support the special leagues out of the box since it'll likely release after WBC

It is a little awkward to ask people to support this new config then swap over to the 9.0.0 configs right after

WardBrian:
> I think the bigger issue is https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/pull/674 is enforcing a teams enum which won't work for the national teams

I'm not actually _enforcing_ that schema any more than the current one. You'll get a yellow squiggle underline in some editors, but it doesn't actually prevent you from putting `"United States"` in the teams list.

---

Within the current config, I think that this kind of thing is not that big a deal to just enable by default. Just based on interacting in the discord etc, it seems like a pretty big majority of users have `'only_preferred': true`, so they wouldn't even see a difference. 

As an aside, we definitely should have Olympics and WBC as different options if we can't actually make the distinction between them, or someone will ask "Why am I getting the olympic games, my config only says WBC"

ty-porter:
Alright, I'm fine with waiting until after 9.0 to deal with it, if we ever need to.

I'll open a separate PR and extract the datetime helper, that's really annoying me

PR #679, opened 2026-03-11
Better path support for configs

ty-porter:
Running the project from outside the project directory currently breaks because it can't find the right config schemas

This updates the path lookups to be relative to project root

* Fixes the bug with starting the project from an arbitrary location
* maintains support for the `--config` option, and better handles relative paths. Previously only supported child paths of project root, now you can put configs anywhere. `.json` file extension is optional
* Centralizes the directories in case we need to load files in the future. Updates a few tests to use them.
* Fixes a secondary bug where typos in the weather location caused a crash

### DIFF
--- a/data/config/__init__.py
+++ b/data/config/__init__.py
@@ -8,6 +8,7 @@ import debug
 from data import status
 from data.config.color import Color
 from data.config.layout import Layout
+from data.paths import *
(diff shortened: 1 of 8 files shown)


ty-porter:
Will reopen against `dev`

PR #686, opened 2026-03-15
Drop support for Python 3.9

ty-porter:
Python 3.9 is EOL as of 10-25.

Potentially disruptive since this project typically runs on system Python as part of Pi OS distro.

At minimum a minor version bump?


WardBrian (approved):

WardBrian:
> At minimum a minor version bump?

Could always keep this unmerged for a bit and roll it into 9.0

PR #688, opened 2026-03-18
Move team name settings to layout file

### LINKED ISSUE #596: 9.0: Move team name settings to coordinates
Right now config.json stores both `full_team_names` and `short_team_names_for_runs_hits` settings, but these should really be in the coordinates file as they are largely layout/size-specific decisions

WardBrian:
Closes #596

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 </a>
 
 ---------------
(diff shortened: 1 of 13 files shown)


WardBrian:
Closing in preference for #689

PR #689, opened 2026-03-18
Refactor configs that belong to layout

### LINKED ISSUE #596: 9.0: Move team name settings to coordinates

ty-porter:

### DIFF
--- a/data/config/__init__.py
+++ b/data/config/__init__.py
@@ -62,14 +62,12 @@ class Config:
         self.weather_apikey = json["weather"]["apikey"]
         self.weather_location = json["weather"]["location"]
         self.weather_metric_units = json["weather"]["metric_units"]
+        self.pregame_weather = json["weather"]["pregame"]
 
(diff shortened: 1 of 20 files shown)


data/config/__init__.py
```diff
@@ -105,6 +103,9 @@ def __init__(self, config_path, width, height):
         self.check_delay()
         self.check_api_refresh_rate()
 
+        # Set up update delay parameter
+        self.sync_amount = self.sync_delay_seconds // self.api_refresh_rate
```
WardBrian: probably want to round up here, see my similar logic in https://github.com/WardBrian/mlb-led-scoreboard/blob/rotation-rework-config/data/schedule.py#L25-L27

WardBrian (approved):

PR #696, opened 2026-03-25
Add schemas for config generator

ty-porter:
Adds schemas and a new `schemas/` module for all the configurations we have. Adds a `*.exmple.json` generator, a workflow to test the examples are clean, and updates some tests.

There's now a pipeline:
* Devs update a schema
* PR can't merge until updating the examples against the schema with `python -m schemas --overwrite`
* Examples now should _never_ be manually edited, even by devs.
* `cp <file>.example.json <file>.json` still the right way to edit custom files.

New `schemas` module CLI:

```sh
Usage:
  python -m schemas                    # regenerate all example files
  python -m schemas --overwrite        # overwrite existing files
  python -m schemas --check            # check if a file would be changed
  python -m schemas --schema schemas/config.schema.json --output config.example.json
```

The CI pipeline runs `python -m schemas --check` and blocks merge until examples are clean

The schemas are optimized for the config editor UI: https://mlb-led-scoreboard.github.io/?version=config-gen-schemas

There's a fairly major quirk here that doesn't follow JSON schema specification -- `$schema` is both a JSON schema keyword and schema association keyword in some editors. The solution is to specify the latter as `x-schema` and have both this repo and the frontend project know to inject that value as `$schema`. If we don't do this some parsers can try to follow the value as if it were a URI and crash.


coordinates/wxhy.schema.json
```diff
@@ -0,0 +1,403 @@
...
+      },
+      "additionalProperties": false
+    },
+    "batter_count": {
+      "type": "object",
+      "properties": {
+        "x":             { "type": "number" },
+        "y":             { "type": "number" },
+        "nohit":         { "$ref": "#/$defs/xy" },
+        "perfect_game":  { "$ref": "#/$defs/xy" }
```
WardBrian: I think we need to think harder about how `font`, `warmup`, `nohit`, and `perfect_game` work here. Basically all of them are valid in all places, so it might make sense to define a `defs/batter_count` and then have the actual key in the top-level object be

```
"$ref": "#/$defs/batter_count",
"additionalProperties" : { "$ref": "#/$defs/batter_count", }
``` 

(or similar)

and then, I guess, do that for every key...

ty-porter: I generated some self-referencing schemas here. These pass online validators I tried.

The form builders I tried (including Jedison) _really_ dislike them, I assume this is causing infinite recursion. I think it might be possible to expand out all the references for each key, but that would make the schema massive and pretty hard to maintain...

WardBrian: Jedison can expand the schema for us (my demo is already doing that https://github.com/WardBrian/mlb-led-scoreboard/blob/gh-pages/config.html#L126-L134) as long as it isn't infinite, which I don't think it needs to be, you might just need another layer of indirection (classic, I know)

Instead of 
```
    "play_text": {
      "type": "object",
      "properties": {
        "x":            { "type": "number" },
        "y":            { "type": "number" },
        "font_name":    { "$ref": "#/$defs/font_name" },
        "desc_length":  { "$ref": "#/$defs/desc_length" },
        "enabled":      { "type": "boolean" },
        "warmup":       { "$ref": "#/$defs/play_text" },
        "nohit":        { "$ref": "#/$defs/play_text" },
        "perfect_game": { "$ref": "#/$defs/play_text" }
      },
      "additionalProperties": false
    }
```

Try something like 

```json
    "play_text_base": {
      "type": "object",
      "properties": {
        "x":            { "type": "number" },
        "y":            { "type": "number" },
        "font_name":    { "$ref": "#/$defs/font_name" },
        "desc_length":  { "$ref": "#/$defs/desc_length" },
        "enabled":      { "type": "boolean" },
      },
      "additionalProperties": false
    },
  "play_text" : {
     "$ref": "#/$defs/play_text_base",
     "addtionalProperties": {  "$ref": "#/$defs/play_text_base"  }
}
```

(I think we could also use patternProperties to enforce that they actually be the required names like `nohit`)

schemas/coordinates/w64h64.schema.json
```diff
@@ -0,0 +1,541 @@
...
+    "network": {
+      "type": "object",
+      "properties": {
+        "font_name":  { "$ref": "#/$defs/font_name" },
+        "background": { "$ref": "#/$defs/rect" },
+        "text":       { "$ref": "#/$defs/point" }
+      }
+    }
+  },
+  "$defs": {
```
WardBrian: We could move these defs to a common file, right?

ty-porter: I think we can. For now I inlined this to get the UI working since it's a lot simpler

I'll probably have to write something on the UI side that complies with schema bundling, unless Jedison handles this for us: https://json-schema.org/understanding-json-schema/structuring#bundling

I also noticed I'm using `$schema` wrong, I think we need to change that to the 2020-12 dialect string, and add `$id` in most cases with the relative path

requirements.txt
WardBrian, line 1: suggestion: make a seperate requirements.txt in `schemas/`, install that in CI, to avoid the base project picking up any more deps?

schemas/config.schema.json
```diff
@@ -445,11 +498,27 @@
     "color": {
       "type": "object",
       "properties": {
-        "r": { "type": "number", "minimum": 0, "maximum": 255 },
-        "g": { "type": "number", "minimum": 0, "maximum": 255 },
-        "b": { "type": "number", "minimum": 0, "maximum": 255 }
+        "r": {
```
WardBrian: I think we can remove this color key now

schemas/colors/teams.schema.json
```diff
@@ -0,0 +1,311 @@
...
+    "mex": { "$ref": "#/$defs/team_colors" },
+    "nca": { "$ref": "#/$defs/team_colors" },
+    "ned": { "$ref": "#/$defs/team_colors" },
+    "pan": { "$ref": "#/$defs/team_colors" },
+    "pr":  { "$ref": "#/$defs/team_colors" },
+    "usa": { "$ref": "#/$defs/team_colors" },
+    "ven": { "$ref": "#/$defs/team_colors" },
+    "kor": { "$ref": "#/$defs/team_colors" },
+    "default": { "$ref": "#/$defs/team_colors" }
+  },
```
WardBrian, line 75: I think we should set additionalProperties to also be ` "$ref": "#/$defs/team_colors"`, in case someone wanted to add e.g. `NEU` to get colors for the once-a-year Red Sox/Northeastern game

schemas/colors/teams.schema.json
```diff
@@ -0,0 +1,311 @@
...
+      "required": ["r", "g", "b"],
+      "properties": {
+        "r": { "type": "number", "minimum": 0, "maximum": 255 },
+        "g": { "type": "number", "minimum": 0, "maximum": 255 },
+        "b": { "type": "number", "minimum": 0, "maximum": 255 }
+      },
+      "additionalProperties": false,
+      "x-format": "custom-color",
+      "x-grid": { "columns": 4 }
+    },
```
WardBrian: Move to a common file?

schemas/colors/scoreboard.schema.json
```diff
@@ -0,0 +1,212 @@
...
+      "type": "object",
+      "x-format": "grid",
+      "properties": {
+        "background": { "$ref": "#/$defs/color", "default": {"r": 7, "g": 14, "b": 25} }
+      },
+      "additionalProperties": false
+    }
+  },
+  "$defs": {
+    "color": {
```
WardBrian: Move to a common file?

schemas/coordinates/w64h32.schema.json
```diff
@@ -0,0 +1,452 @@
...
+      "properties": {
+        "1": { "$ref": "#/$defs/out_pos" },
+        "2": { "$ref": "#/$defs/out_pos" },
+        "3": { "$ref": "#/$defs/out_pos" }
+      }
+    },
+    "atbat": {
+      "type": "object",
+      "properties": {
+        "batter": {
```
WardBrian: my personal config has a 'nohit' key in the batter object. I think in general any key can use it, actually

schemas/coordinates/w64h32.schema.json
```diff
@@ -0,0 +1,452 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "64x32 Scoreboard Coordinates",
+  "x-indent": "  ",
+  "type": "object",
+  "required": ["format"],
+  "properties": {
```
WardBrian, line 7: Couldn't everything besides x-schema and defaults also be in 'common' as `$defs/config` or similar?

WardBrian (commented):
This looks great. Comments on the schemas here, plus two thoughts on the website:

1. it would be nice to make the text box editable so you can paste in an existing config to edit
2. I do think some option that you can cut/paste into a terminal is useful (e.g. https://github.com/WardBrian/mlb-led-scoreboard/blob/gh-pages/config.html#L117-L120)

schemas/coordinates/w64h32.schema.json
ty-porter: I've bashed against this for a couple days now, I'm not sure why but it looks like Jedison has a memory leak or something here. For all the keys I should be able to do something like:

```json
"atbat": {
      "type": "object",
      "properties": {
        "batter": {
          "type": "object",
          "properties": {
            "x":         { "type": "integer" },
            "y":         { "type": "integer" },
            "width":     { "type": "integer" },
            "font_name": { "$ref": "#/$defs/font_name" },
            "offset":    { "type": "integer" }
            "nohit":     { "$ref": "#/$defs/atbat/batter" },
            < other states >
          },
          "required": ["x", "y", "width"]
        },
```

Such approach takes a _very_ long time to render or crashes the tab entirely for OoM. Very similar symptoms to https://github.com/germanbisurgi/jedison/issues/32

This doesn't have `additionalProperties = false`, so for now you can do this manually:

image

ty-porter: Maybe we could revisit this feature altogether 😄

WardBrian: I think the feature is useful, but I'm also fine with it being something you need to crack open a text editor for rather than it being nice and in Jedison. As long as I don't get red squiggly lines when I do open that editor!

WardBrian (approved):
I think this is ready to fly. I am traveling the next couple days but I might look into getting the nohit/perfect_game stuff working later, I don't think it's a big enough feature to be a blocker.

Thanks @ty-porter!

As an aside, if we do start allowing arbitrary plugins, the smallest change would just be allowing additionalProperties at the top level of all these, or perhaps a "plugins" key that can have arbitrary sub properties.

PR #698, opened 2026-03-30
Add plugin system for secondary screens

### LINKED ISSUE #695: Ability to add more non-game screens as plugins
I mentioned this in #667 but it's worth discussing separately, and deciding if we want to include it in 9.0

Essentially, things like the standings screen are very easy to imagine more of. Baseball related you could have a league HR leaders screen, non-baseball related we already have examples of upcoming subway times or wait times at Disney world...

From a code perspective, all this looks like is:

A name (like "standings")

A class that can be constructed from a configuration file and that has a .update() function

A render function that accepts that data object and the layout. Maybe it also should accept the matrix from the outside, and a scrolling text integer to be safe

I think the hard part is deciding how we'd like these to interact with the config/layout files: should we force them to use different files, do we just have some extra section in our config class for these, etc, etc.

WardBrian:
Would close #695 

This adds a new library (`bullpen`) that provides an API for defining a plugin, which we treat as the 4-tuple `(name, config_class, data_class, render_class)`.

Any plugin so-defined and registered under `[project.entry-points.'bullpen.mlbled.plugin']` in its pyproject.toml becomes an option for a secondary screen. The existing standings and news screens have been ported to this system, now living in the `news/` and `standings/` folders rather than mixed in with the rest of the code.

Before merging, this needs:
- A few more examples (I want to port https://github.com/WardBrian/mta-board)
- a lot of doc, at least in the form of a readme in `bullpen/`
- updates to the schemas for allowing the `plugins` key where needed.


bullpen/src/bullpen/api/config.py
```diff
@@ -0,0 +1,54 @@
+import abc
+import datetime
+from typing import TYPE_CHECKING, Any, Protocol, Union
+
+if TYPE_CHECKING:
+    from RGBMatrixEmulator.graphics import Color as GraphicsColor
+
+
+class MLBConfig(Protocol):
```
ty-porter, line 9: This feels like we should separate concerns here. It would be cool for bullpen to be a framework agnostic to whatever project is using it. 

We could extract this into a package that abstracts away all the complexity of managing the plugin state machine. That said it's so far down the road I have no issues as-is, just something to think about

data/config/__init__.py
```diff
@@ -189,8 +171,26 @@ def parse_today(self):
...
+
     def screen_time_at_priority(self, screen: str, priority: int) -> int:
-        return self.rotation_screen_rules.get(screen, {}).get(priority, 0)
+        return self.rotation_screen_rules.get(priority, {}).get(screen, 0)
+
+    def for_plugin(self, plugin_name: str) -> MLBConfig:
+
+        match plugin_name:
+            case "news":
+                # for legacy reasons, we let this plugin have two separate config names
```
ty-porter, line 186: Is there a reason we can't break this convention?

I assume this is because we show weather on news and have a dedicated weather screen?

data/__init__.py
```diff
@@ -1,61 +1,37 @@
...
-
-    def refresh_schedule(self):
+    def refresh_schedule(self) -> None:
         self.__process_network_status(self.schedule.update())
 
-    def __process_network_status(self, status):
+    def refresh_plugins(self) -> None:
+        self.__process_network_status(UpdateStatus.merge(plugin.update() for plugin in self.plugin_data.values()))
+
+    def __process_network_status(self, status) -> None:
```
ty-porter, line 46: Nitpick but I don't see much advantage over the following (less indirection):

```suggestion
    def refresh_plugins(self) -> None:
       status = UpdateStatus.merge(plugin.update() for plugin in self.plugin_data.values())
```

data/plugins.py
```diff
@@ -0,0 +1,37 @@
...
+            )
+            cfg_class, data_class, renderer_class = plugin()
+            cfg = cfg_class(config.for_plugin(name))
+            data = data_class(cfg)
+            renderer = renderer_class(cfg, config.layout, config.scoreboard_colors)
+        except Exception as e:
+            if config.debug:
+                raise ValueError(f"Error loading plugin {name} from {entry_point.module}") from e
+            else:
+                debug.warning("Error loading plugin %s from %s: %s", name, entry_point.module, str(e))
```
ty-porter: seems like something we should never swallow in a warning

news/src/mlb_led_scoreboard_news/headlines.py
```diff
@@ -110,23 +113,23 @@ def __init__(self, config, year):
     def update(self, force=False) -> UpdateStatus:
         status = UpdateStatus.SUCCESS
         if force or self.__should_update():
-            debug.log("Headlines should update!")
+            LOGGER.debug("Headlines should update!")
```
ty-porter, line 116: 👍🏻 on this change. I'd love it if we transitioned to this pattern across the whole project rather than the weird custom `debug` module

ty-porter (commented):
This looks great, exciting to finally clean up some of the spaghetti. Left a few minor comments.

data/config/__init__.py
WardBrian: This workaround is just to preserve the existing config structure. If we wanted to, we could change the existing config to look like 

```json5
"plugins": {
  "news": {
    "teams": [
      "Cubs"
    ],
    "traderumors": true,
    "mlb_news": true,
    "countdowns": true,
    "date": true,
    "date_format": "%A, %B %-d"
    "owm_apikey": "YOUR_API_KEY_HERE",
    "location": "Chicago,il,us",
    "metric_units": false,
    "pregame": true // this one is still a little odd...
  },
  "standings": {
    "divisions": [
      "NL Central",
      "NL Wild Card"
    ]
  }
}
```

bullpen/src/bullpen/api/config.py
WardBrian: The underlying plugins do need access to some information from the overarching board config, though. The name could be different, but I think something like this does need to be around

news/src/mlb_led_scoreboard_news/headlines.py
WardBrian: Done 😄

bullpen/src/bullpen/api/config.py
ty-porter: No worries. This is pie-in-the-sky thinking anyway

ty-porter (approved):

PR #701, opened 2026-04-02
Add dev requirements black & mypy

ty-porter:
Adds lint and typechecking CI workflows + dev requirements. Neither of these workflows are blocking, so we can evaluate these as needed. Lint checks run only on changed files.


WardBrian:
At a minimum https://github.com/WardBrian/mlb-led-scoreboard/blob/96bf29bbb4700c19634af56aca89a48c4401f753/driver/__init__.py#L7-L8 should help with some of the type errors. The trickery we are doing in `driver` is not well recognized.

PR #702, opened 2026-04-02
Allow reading from config.json for matrix command line flags

### LINKED ISSUE #700: Allow storing matrix config options in config.json
As an alternative to the command line options. If both are present, I'd say the command line wins, but it would also be reasonable to error out

ty-porter:
Closes #700 

Adds ability to read from a config file `matrix` subsection for the CLI flags. This is more complex than it sounds due to 1) a circular dependency on CLI read vs. config instantiation and 2) tests also reading command line flags implicitly

We read flags in 2 passes. The first pass picks up `--config` if it exists. This lets a new config object read the values in `matrix` and pass it back to the parser which uses those values as default.

The priority of flags is user-specified CLI flags > JSON config flags > argparse defaults

updated and the config validator is expanded to completely ignore the `matrix` subkeys when removing keys. It still adds the section overall if it is missing.


schemas/config.schema.json
```diff
@@ -6,6 +6,9 @@
   "required": [
     "format"
   ],
+  "patternProperties": {
+    "^\\$schema$": { "type": "string" }
```
ty-porter, line 10: All the tooling around this is brittle because the keywords are colliding... This is the best solution I could come up with.

data/config/__init__.py
```diff
@@ -288,7 +294,15 @@ def __get_layout(self, width, height):
...
-        return isinstance(other, Config) and vars(self) == vars(other)
+        self_keys = { k: v for k, v in vars(self).items() if k != 'matrix_options' }
+        other_keys = { k: v for k, v in vars(other).items() if k != 'matrix_options' }
+
+        keys_match = self_keys == other_keys
+
+        # Spot check matrix options. These don't need strict equality, for config we only care about size.
+        options_match = self.matrix_options.cols == other.matrix_options.cols and self.matrix_options.rows == other.matrix_options.rows
+
+        return isinstance(other, Config) and keys_match and options_match
```
WardBrian: We should move the isinstance check to the top of the function to avoid an issue with calling `other.matrix_options`

data/config/__init__.py
```diff
@@ -334,3 +348,59 @@ def _screen_rules_from_json(json) -> tuple[list[GameScreen], list[TimeRule], Map
             )
 
     return game_rules, time_rules, screen_rules
+
+def _matrix_options(json, clargs):
+    args = Namespace(**json)
+
+    for flag, setting in vars(clargs).items():
+        if flag in sys.argv or flag not in args:
```
WardBrian: I find this logic quite confusing -- is it just because we want to be able to tell whether something was defaulted or not?

data/config/__init__.py
```diff
@@ -334,3 +348,59 @@ def _screen_rules_from_json(json) -> tuple[list[GameScreen], list[TimeRule], Map
...
+            args.__setattr__(flag, setting)
```
WardBrian:
```suggestion
            setattr(args, flag, setting)
```

data/config/__init__.py
WardBrian: The other way to do this, which might be easier, is to do it in two steps: 

argparse just the `--config` flag, open that, and pass `json['matrix']` to a function that uses the values in that as the defaults in a second call to `argparse`. That way you don't need to try manually merging anything

ty-porter: Yeah it's a little clunky but I think it's the right approach without dealing with parsing args in multiple passes. This is solving for the chicken-and-egg problem of having CLI flags and config dependent on each other.

argparse defaults show up in a `Namespace` so it's difficult to see which was actually passed on the command line without looking at ARGV. We want to use JSON to overwrite anything EXCEPT command line args. I also didn't want to duplicate the CLI flags into the `matrix` config section so all the keys there are optional

So what this is doing is prioritizing user CLI flags, JSON config, argparse defaults in that order

WardBrian: Does `flag in sys.argv` really work? I'd expect `flag` to be something like `led-cols` while sys.argv obviously contains something like `--led-cols=32`

ty-porter: You're right, mistake on my part, I did need to add a normalization step and clean this up a bit.

driver/__init__.py
```diff
@@ -1,14 +1,14 @@
 import sys
 
-from utils import args
+from cli import ScoreboardCLI
```
WardBrian:
```suggestion

```

data/config/__init__.py
```diff
@@ -286,9 +296,73 @@ def __get_layout(self, width, height):
...
             return new_layout
         return reference_layout
+    
+    def __matrix_options(self, json):
+        cli = ScoreboardCLI()
+        args = cli.canonical_arguments(json)
+
+        self.emulated = args.emulated
+
+        options = RGBMatrixOptions()
```
WardBrian, line 309: Q: should ScoreboardCLI.canonical_arguments() just return a `tuple[bool, RGBMatrixOptions]`? 

Also, note that you _are_ parsing the command line twice...

cli.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        defaults = vars(self.default_arguments())
+        clargs = vars(self.arguments())
+
+        argv_flags = {self.__normalize(f) for f in sys.argv}
+
+        config_flags = {flag: json_data[flag] for flag in defaults if flag in json_data}
+        explicit_flags = {flag: clargs[flag] for flag in defaults if flag in argv_flags}
+
+        return argparse.Namespace(**(defaults | config_flags | explicit_flags))
```
WardBrian: Since this method parses the args again anyway, I really think it's worth just writing `__make_parser` such that it accepts a `defaults={}`. 
Then for e.g. `--led-cols`, `default=defaults.get("led_cols", 32)`

Then the control flow looks like
- call make_parser with `{}`,  extracting only `config`
- call make_parser with `config['matrix']`, convert to RGBMatrixOptions

That way we're not having to figure out after-the-fact which things where in sys.argv or not, what the defaults should be, etc, etc

WardBrian (commented):
edit: didn't mean to post a top level comment here

WardBrian (approved):
I'm not able to try it out locally atm, but reading over the changes I think this is good to go!

PR #719, opened 2026-04-13
Bump rgbmatrix version

WardBrian:


ty-porter (approved):

ty-porter:
That test should be fixed after #718

PR #720, opened 2026-04-15
New CC colors for 2026

WardBrian:


ty-porter (approved):
The diff UI is great!

Looks like Atlanta is pretty low contrast

PR #723, opened 2026-04-17
Add hide_record_on_high_line_score coordinates option

ChickenSupreme:
My apologies if I didn't format this PR correctly. I am still very unfamiliar with git.

This PR adds a "hide_record_on_high_line_score" option to the coordinates files. I have been manually porting this change along for years on my 64x32 board because I want to display team records and there's just barely enough room with short team names when the line score doesn't contain any double digit entries.

Some minor caveats: I believe the record's closing bracket will just barely touch the line score when the record has a three digit number (e.g. 100-62) on my 64x32 board, i.e. there's no space between them. Also, this will prematurely hide records even when there IS enough room when the records are small early in the season and only contain single digit numbers.

One other unrelated note: When I was duplicating the can_use_full_team_names function I noticed that it looks like it might be pulling the shorten_team_name_on_high_line_score option from the wrong config item? This isn't an option I use so I didn't test it but it looked wrong to me when comparing with the example json files.

Thank you for considering this PR.

### DIFF
--- a/coordinates/w128h32.example.json
+++ b/coordinates/w128h32.example.json
@@ -318,7 +318,8 @@
         "font_name": "4x6",
         "x": 19,
         "y": 17
-      }
+      },
+      "hide_record_on_high_line_score": false
     },
     "line_score": {
       "away": {
--- a/coordinates/w128h64.example.json
+++ b/coordinates/w128h64.example.json
@@ -296,7 +296,8 @@
       "home": {
         "x": 30,
         "y": 29
-      }
+      },
+      "hide_record_on_high_line_score": false
     },
     "line_score": {
       "away": {
--- a/coordinates/w192h64.example.json
+++ b/coordinates/w192h64.example.json
@@ -296,7 +296,8 @@
       "home": {
         "x": 30,
         "y": 29
-      }
+      },
+      "hide_record_on_high_line_score": false
     },
     "line_score": {
       "away": {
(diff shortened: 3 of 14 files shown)


ChickenSupreme:
I'm not sure where the extra commit came from, maybe I picked the wrong branch? It said to submit against the dev branch but I think I made my commits to master.

Also I should note that I set the defaults in the example json files to true for any board of width 64 or less and false for any board of width greater than 64. I left the record display disabled for all of them though so that would only take effect if someone manually enables records on their board. I only have a 64x32 board so I don't actually know if those defaults make sense - maybe you'd prefer if they were all the same.

ty-porter (commented):
Thank you for this PR, this looks mostly great!

I just ran the build checks on your PR, you may see a failure here. This is a new process we put in place in v9.0.0 where we edit `schemas/coordinates/wxhy.schema.json` files and then generate the examples from the schema. I'm just now realizing we forgot to document this, so if you aren't comfortable making that change, let me know and we can have one of the core team can update it for you later. We won't let that block this PR.

---

> I'm not sure where the extra commit came from, maybe I picked the wrong branch? It said to submit against the dev branch but I think I made my commits to master.

That's probably what happened. It's a minor problem and this project isn't too picky about git history so I don't think you need to do anything. The code changes look fine.

ChickenSupreme:
I updated all the schema files with the new option.

ty-porter (approved):

PR #732, opened 2026-04-25
Add dynamic uniform type detection via colors/teams.json

jc214809:
Any non-standard key in a team's color object (beyond home/text/accent) is
automatically registered as a uniform type. The MLB API detection string is
derived from the key name (snake_case -> Title Case) with case-insensitive
matching, so city_connect matches "City Connect 2.0 Jersey" and cincy matches
"CINCY Jersey" without any hardcoding. Adds Reds CINCY alternate jersey colors
to colors/teams.example.json and updates schema to allow arbitrary uniform
type sub-objects.


validate_config.py
```diff
@@ -10,6 +11,25 @@ def make_modifier(mod):
     return MODIFIER + mod
 
 
+def _colors_ignored_keys():
```
ty-porter: This subverts how the validator works and seems fairly flaky -- this validator is complicated so I'm worried that changing how it works based on user input is a potential minefield. There's also an upsert feature which checks if there's a missing key and adds it if it's not there, and removes things that aren't needed any more. For better or worse, it's one of the only ways we have to automatically migrate configs when versions change.

The special `ignore_keys` setting overrides both of those migration features, so this is a little unsafe.

---

What if we added a dict of special uniforms:

```json
"cin": { 
  "home": {
    "r": 1,
    "g": 2,
    "b": 3
  },
  "text": {
    "r": 1,
    "g": 2,
    "b": 3
  },
  "accent": {
    "r": 1,
    "g": 2,
    "b": 3
  },
  "special_uniforms": {
    "city_connect": {
      "r": 1,
      "g": 2,
      "b": 3
    },
    "cincy": {
      "r": 1,
      "g": 2,
      "b": 3
    }
  }
}
```

What this lets us do is mark `special_uniforms` ignored:
```python
"ignored_keys": [
  "special_uniforms" + make_modifier(IGNORE_SUBPATHS),
  "plugins" + make_modifier(IGNORE_SUBPATHS)
]
```

ty-porter: Unfortunately this strategy will expand the scope of this PR to touch every single team's colors entry and is also a breaking change, anyone with an existing custom `city_connect` setup won't get migrated automatically.

I'm personally OK with breaking those custom colors

colors/teams.example.json
```diff
@@ -237,6 +237,23 @@
         "g": 239,
         "b": 237
       }
+    },
+    "cincy": {
```
ty-porter: This will probably have to get added to the schema as a default value

ty-porter (commented):
This is a great feature and future-proofs new team color additions. I think we need to make sure it'll work more holistically with our existing tooling.

I left some suggestions to cover that.

data/config/__init__.py
```diff
@@ -24,6 +24,19 @@
...
+
+_STANDARD_COLOR_KEYS = {"home", "text", "accent"}
+
+
+def _extract_uniform_types(teams_json: dict) -> dict:
+    uniform_types = {}
+    for team_data in teams_json.values():
+        if not isinstance(team_data, dict):
+            continue
+        for key in team_data.get("special_uniforms", {}):
```
WardBrian, line 36: I think I'd vote to keep them top level, and just have it be

```python
for key in team_data:
  if key in ('r', 'g', 'b'): continue
  # same as current
```

That way this PR doesn't need to change any of the jsons, and it doesn't break anyone's existing custom config

data/config/__init__.py
```diff
@@ -24,6 +24,19 @@
...
+            if key not in uniform_types:
+                uniform_types[key] = key.replace("_", " ").title()
```
WardBrian, line 38: rather than call .title now and .lower in uniforms.py, why not just .lower them here?

data/uniforms.py
```diff
@@ -8,20 +8,19 @@
...
-
-# maybe one day we will also want to support e.g throwback
-SPECIAL_UNIFORMS = {CITY_CONNECT: lambda uniformName: "City Connect" in uniformName}
-
 
 # separate API call and not something we expect to change, so we don't do
 # this as part of the Game data updates
 class Uniforms:
-    def __init__(self, game_id):
+    def __init__(self, game_id, uniform_types: dict):
```
WardBrian, line 15: I wonder if, instead of passing the uniform data through game from schedule.py, if we should just update the `home_special_uniform`/`away_special_uniform` functions to accept the types dict as an argument at lookup time?

WardBrian (commented):
Overall this looks pretty good to me, mainly a few what-if questions

ty-porter:
Merging this as-is, I have spare time to tweak on this and implement @WardBrian's suggestions. After further consideration I agree, we should put the uniforms top-level for each team for backwards compatibility.

PR #739, opened 2026-05-15
Refactor special uniforms

ty-porter:
Refactors around #732

* Moved new special uniforms from `special_uniforms` back to team top-level key. This prevents us from breaking existing configs.
* Removed the `ignored_keys` entry in config validation. We want all custom configs to have these keys, and this change makes sure everyone gets it. My memory is fuzzy, perhaps in the past we "turned off" City Connect support by making City Connect keys optional and allowing users to delete them? If so, we should move that to a config instead (outside the scope of this PR).
* Added some tests around the new functionality, specifically the `uniforms` module. Made a slight refactor here to avoid wonky syntax inviting a [loop-over-closure bug](https://medium.com/skiller-whale/late-binding-variables-its-a-trap-c17af980164f) if we ever refactored that code
* Ran the linter and typechecker (source of most failures in the original PR)
* Added the new `cincy` key to the schema and re-ran schema generation. This was done by hand in the original PR


WardBrian:
> Team colors changed in this PR! Preview: [Team Colors](https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/actions/runs/25938484810/artifacts/7025400217)

Something about this seems to have broken here

Edit: I think just a missing reference https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/35787ad2dd12c046802ece230866557820221913/data/scoreboard/team.py#L31

WardBrian:
I still wonder about https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/pull/732#discussion_r3177223670, if it would be cleaner to avoid needing to add another argument to Game.from_scheduled

ty-porter:
> I still wonder about [#732 (comment)](https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/pull/732#discussion_r3177223670), if it would be cleaner to avoid needing to add another argument to Game.from_scheduled

The available uniforms are part of config, but the `Scoreboard` data object currently has no context of config:

https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/394834484969b03ae831606d11795fc286963710/data/scoreboard/__init__.py#L21-L39

I think something is going to have to accept the list of uniforms. In my opinion the `Game` object is the ideal place to do it since uniform lookups require game IDs. We already pass around pieces of config to the game, so I think the easiest solution is just give it the config directly.

Pushed a commit with this change, but we can revert it if we decide we don't like it.

32452bc041339bce1a0efccb6d541af35128df15

WardBrian (approved):
Looks good!

PR #740, opened 2026-05-17
Add a config value to disable special uniform checks

ty-porter:


WardBrian:
Couldn't someone just delete those keys or set them to the same colors as the normal team?

ty-porter:
> Couldn't someone just delete those keys or set them to the same colors as the normal team?

They could set them the same as the normal team (would be a lot of work potentially), but deleting the keys will just have the config validator re-add them next time they a reinstall happens.

WardBrian:
ah, because we can't tell the difference between a key that was deleted and one that never existed. Except, I suppose, now we could based on the format field

PR #743, opened 2026-05-20
Add plugin priority handling and update data initialization

jc214809:
- Enhanced Data and Schedule classes to accept plugin renderers and data.
- Updated configuration schema to include plugin screen priorities.
- Modified _screen_rules_from_json to handle new plugin priority rules.


data/schedule.py
```diff
@@ -112,6 +114,13 @@ def __filter_games(self, all_games: list) -> tuple[int, list]:
             if priority:
                 highest = max(highest, priority)
 
+        for plugin_name, priority in self.config.rotation_plugin_priority_rules.items():
+            renderer = self._plugin_renderers.get(plugin_name)
+            data = self._plugin_data.get(plugin_name)
+            if renderer is not None and data is not None and renderer.can_render(data):
```
WardBrian: I think we'd probably want a separate function than can_render. I don't love storing references to the renderers on the data thread, and it's kind of a separate concern anyway. I will try to think about what the desirable API is here. I think it would make sense to let plugins that want to define a priority take in some sort of information for filtering, like the LL plugin could support a `'requires':'park_open'`

data/config/__init__.py
```diff
@@ -400,8 +398,18 @@ def _screen_rules_from_json(json) -> tuple[list[GameScreen], list[TimeRule], Map
...
-                screen_rules[priority][rule_json["kind"]] = rule_json["seconds"]
+            kind = rule_json["kind"]
+            if "priority" in rule_json:
+                priority = rule_json["priority"]
+                if priority == 0:
+                    raise ValueError(
+                        f"Invalid plugin rule in config, priority cannot be 0. Rule: {rule_json}"
+                    )
+                plugin_priority_rules[kind] = priority
+                screen_rules[priority][kind] = rule_json["seconds"]
```
WardBrian: currently the 'primary' rules don't have a seconds field. I'd want to think about what this means, because it doesn't really gel with my idea of how the priorities would work

WardBrian (changes requested):
I think we can get something like this in an upcoming version, but I want to take some time to think about how it would be best fit in. Something like this makes it tempting to think of games themselves as a kind of plugin, but I'm not sure that is actually desirable

data/schedule.py
jc214809: 'requires':'park_open' I like this idea!

PR #752, opened 2026-06-01
Add MLB.com recap and preview blurbs to game scroll text

twfarley:
- Game._fetch_game_content() pulls the MLB editorial section from the game_content API endpoint, cached 5 min per game
- game_recap_blurb() and game_preview_blurb() build "headline — subhead" strings, returning "" when content has not yet been published (typical for the first ~10-30 minutes after a game ends)
- Postgame: recap blurb is appended to the W/L scrolling line
- Pregame: preview blurb is appended to the pitchers/broadcasts/weather scrolling line
- Stop refreshing game data once abstractGameState is "Final" to avoid pointless API polling on completed games

### DIFF
--- a/data/game.py
+++ b/data/game.py
@@ -9,6 +9,7 @@ from data import teams
 from bullpen.api import UpdateStatus
 from data.utils.circular_queue import CircularQueue
 from data.uniforms import Uniforms
+from data.blurbs import Blurbs
 from data.scoreboard import Scoreboard
 from data.scoreboard.postgame import Postgame
 from data.scoreboard.pregame import Pregame
@@ -59,6 +60,7 @@ class Game:
         self._api_refresh_rate = config.api_refresh_rate
         self._status: dict[str, Any] = {}
         self._uniform_data = Uniforms(game_id, config.uniform_types)
+        self._blurb_data = Blurbs(game_id)
 
     def update(self, force=False, testing_params={}) -> UpdateStatus:
         if force or self.__should_update():
@@ -91,6 +93,7 @@ class Game:
                         LOGGER.error("Failed to get game status from schedule")
 
                 self._uniform_data.update()
+                self._blurb_data.update()
                 self.print_game_data_debug()
                 return UpdateStatus.SUCCESS
             except:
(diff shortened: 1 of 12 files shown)


data/game.py
```diff
@@ -359,7 +359,40 @@ def current_play_result(self):
...
+            content = statsapi.get(
+                "game_content",
+                {"gamePk": self.game_id},
+                request_kwargs={"headers": data.headers.API_HEADERS},
+            )
+            self._game_content_cache = content
+            self._game_content_cache_time = now
+            return content
+        except Exception:
+            return {}
```
WardBrian: Rather than implement its own caching here, maybe this should be pulled into a small `Blurbs` class with it's own `.update()` and everything, similar to [`uniforms.py`](https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/master/data/uniforms.py)? We currently have a loose organization where each endpoint has its own class, which I think is nice to preserve

twfarley:
I guess it should be noted the the game preview blurbs are pretty rare, but the postgame usually show up within 30 minutes of the game ending.

renderers/games/postgame.py
```diff
@@ -32,6 +32,9 @@ def _render_decision_scroll(canvas, layout, colors, postgame, text_pos, is_playo
     if postgame.save_pitcher:
         scroll_text += " SV: {} ({})".format(postgame.save_pitcher, postgame.save_pitcher_saves)
 
+    if postgame.recap_blurb:
+        scroll_text += "  —  " + postgame.recap_blurb
```
WardBrian: Do all of our pixel fonts have `—`? It might be safer (if typographically incorrect) to use `-`

renderers/games/pregame.py
```diff
@@ -48,6 +48,9 @@ def _render_pregame_info(canvas, layout, colors, pregame: Pregame, probable_star
     if pregame_weather and pregame.pregame_weather:
         pitchers_text += " Weather: " + pregame.pregame_weather
 
+    if pregame.preview_blurb:
```
WardBrian: [question] we have pregame_weather as a config value you can turn on and off, do you think we should for these as well?

WardBrian (commented):
Thanks. I'd be pretty happy to merge this as is, just two small questions

renderers/games/postgame.py
twfarley: Yeah, you're right. Updated to hyphens.

renderers/games/pregame.py
twfarley: Yeah, I think it might get annoying on w32 boards if the blurb is really long. Addressed.

WardBrian (approved):

ty-porter (approved):

WardBrian:
Just saw this for the first time live on my board, gotta say I love it!

PR #758, opened 2026-06-03
Update driver for Pi4 and below compatibility

ty-porter:
~~Reverts the driver SHA bump since there is a known issue with that fails to compile on Raspberry Pi OS for Pi versions < 5~~

Rolls forward driver with latest changes to compile in Raspberry Pi OS for Pi 4 and below.


WardBrian:
Should we bump this to 1ee4f76f2b6061b198bb3a14970bea5cefdfeefc now?

WardBrian (approved):

Issue #705, opened 2026-04-06
Add ABS Challenges to dataset and layouts


tracerrx:
Add remaining ABS challenges for each team to the internal dataset from the API and integrate them into each layout.

<img width="2880" height="2160" alt="Image" src="https://github.com/user-attachments/assets/97a98d69-137c-4722-887d-48fca22d13c2" />

WardBrian:
Interesting that this information is being displayed prominently, while I don't recall the same treatment ever being given to normal replay review challenges on a scorebug

tracerrx:
Thats a valid point... I liked having the ABS challenges up there though...  Maybe because it's "new and shiny"

gotopes:
for whoever might want to implement this:

https://statsapi.mlb.com/api/v1.1/game/(gamePk here)/feed/live   URL call

look for: 

 "absChallenges": {
      "hasChallenges": true,
      "away": {
        "usedSuccessful": 1,
        "usedFailed": 2,
        "remaining": 0
      },
      "home": {
        "usedSuccessful": 0,
        "usedFailed": 1,
        "remaining": 1

If you also want to add mound visits, not that much further away from the ABS data:

   "moundVisits": {
      "away": {
        "used": 2,
        "remaining": 3
      },
      "home": {
        "used": 0,
        "remaining": 5

WardBrian:
#759

PR #759, opened 2026-06-03
ABS indicator

twfarley:
Adds a colon-style indicator to the right of the score on the live-game screen that shows each manager's remaining ABS challenges. Lit (bright yellow) squares are available; dim (mustard) squares are spent. Fills from the bottom up — when one is used, the top square dims first.
   
  - New `Game.abs_challenges_remaining(side)` pulls `gameData.absChallenges.<side>.remaining`; `API_FIELDS` extended to include the field
  - `Scoreboard` exposes `away_abs_challenges` / `home_abs_challenges`, defaulting to 2 when the API hasn't populated the field 
  - New `render_abs_challenges` is called from `MainRenderer` **after** `render_team_banner` so the squares sit on top of the team-background fill; gated by `status.is_live(...)` so it only shows on live games
  - Schema additions:
    - `teams.line_score.show_abs_challenges` (bool, default `true`) — the on/off toggle, sibling of `show_hits_and_errors`
    - `teams.abs_challenges` — per-side `{x, size, squares}` for placement
    - Two new color keys: `abs_challenges.available` (yellow) and `abs_challenges.used` (dim mustard)
  - Per-size defaults added to all 6 example coord files. `w128h64` and `w192h64` shift `line_score.x` slightly left to make room for the 3×3 colon; smaller screens use 2×2 squares.
   
  ## Test plan
  - [x] `python -m unittest discover` — 94 tests pass
  - [x] `black --check`, `mypy`, `python -m schemas --check` all clean
  - [x] Verified in emulator at `w128h64` during a live game — indicator visible, fills from bottom on spend
  - [x] Reviewer: spot-check the smaller layouts (`w32h32`, `w64h32`) for cramped placement


twfarley:
This would address #705

renderers/main.py
```diff
@@ -143,6 +143,11 @@ def __draw_game(self, game: Game):
             show_score=not status.is_pregame(game.status()),
         )
 
+        # ABS challenge indicators sit on top of the team-banner background,
+        # so render them after the banner -- and only while the game is live.
```
ty-porter: Sounds like we want this in `teams.render_team_banner()` between background and any text, otherwise in theory we could draw over team names or scores

🙃 as much as I hate threading more state parameters through those methods

ty-porter (commented):
Very excited for this, happy to help test various sizes if you need

ty-porter (approved):

PR #763, opened 2026-06-09
Add performance profiling support

ty-porter:
I added support for performance profiling to diagnose some RGBME bottlenecks, thought it could be useful here.

image

Interesting that we spend over 90% of our time busy waiting

image

https://github.com/ty-porter/RGBMatrixEmulator/pull/136


WardBrian:
Shame that Tachyon isn’t in the stdlib until 3.15, but cProfiler is probably fine 

is that flame graph for both threads?

ty-porter:
No, this is specific to the render thread. I figured the data thread would be fairly sparse outside of periodic updates, and our producer/consumer architecture would hide issues anyway

WardBrian:
How hard would it be to profile both threads? I'm particularly thinking this would be useful to identify expensive or misbehaving plugins

ty-porter:
I added a separate profiler for the main thread. I think you can _probably_ combine these if we wanted to, but I think for this project keeping them separate is a little clearer.

image

Right off the bat, loading plugins is pretty IO bound

WardBrian (approved):

PR #764, opened 2026-06-10
Add frame pacing for smoother frame transitions

ty-porter:
One more PR related to perf. I came up with this after working #763, but separated it out since they are separate concerns.

Previously we slept after each drawn frame for equal amounts of time regardless of how the frame is drawn. It could cause "jerky" frame updates if one frame unexpectedly takes a long time. This introduces a frame pacing context manager to ensure the time between frames is identical regardless of how long the frame draw routines take. Obviously, if we completely consume the wait time budget, it'll still delay the next frame.

This change is a relatively minor optimization. Our draw routines are currently not very heavy (at least on emulator).


WardBrian:
I love this. Is there necessarily a reason you want it to live in bullpen? I’m not opposed, it just seems like a implementation detail

ty-porter:
I supposed it doesn't have to, it could be moved to the main renderer instead. You're right, bullpen doesn't use this directly

WardBrian (approved):

PR #765, opened 2026-06-28
Config web editor

twfarley:
Adds a "Web Config Editor" subsection (and TOC entry) under Configuration, and
documents the new top-level `sport_ids` option in the config reference. This is
the final piece of the schema-driven configuration workflow built on this
branch; the notes below recap the whole feature for posterity.

## Web config editor (config_editor.py + config_editor_static/)

A dependency-free local web app (stdlib http.server, vanilla JS — nothing to pip
install on a Pi) that turns config editing into point-and-click:

- Schema-driven form. The UI is generated directly from schemas/config.schema.json,
  so every option renders with the right control and its schema description as
  inline help, and new schema options appear automatically:
    * booleans -> checkboxes
    * enums -> dropdowns
    * bounded numbers marked x-format:"range" -> sliders with a live readout
      (scrolling_speed 0-6; api_refresh_rate 15-60s, step 5)
    * arrays of enums -> multi-checkbox "choices" (teams, divisions, leagues)
    * rotation.screens -> a repeatable row editor with a per-row kind dropdown
    * demo_date (false | "YYYY-MM-DD") -> an enable checkbox + native date picker
    * freeform objects (plugins) -> JSON passthrough, preserved verbatim
    * matrix driver options are hidden (x-hidden) but passed through untouched
- Reads custom-or-example, writes custom. Loads config.json when present, else
  config.example.json, and always writes config.json — existing settings are the
  starting point, never bare defaults. The same example-merge + reconcile path as
  the CLI verifier (validate_config.upsert_config) keeps keys in sync.
- Refuses to write a broken config. Every save runs through the board's own
  parser (_screen_rules_from_json, get_team_id); anything that would crash
  Config() at startup — priority 0, unknown team, missing with_priority:0 fallback,
  conflicting time rules — is rejected with the exact error and nothing is written.
- Backs up every save (timestamped config.json.bak-…; now gitignored via *.bak-*).
- Optional Save & Restart. Auto-detects a systemd unit / launchd label matching
  mlb*scoreboard and offers a one-click restart; otherwise saves and tells the
  user to restart manually. Override with --service NAME.

## Leagues & teams

- New `sport_ids` config option (default [1] = MLB), surfaced as the "Leagues"
  selector at the top of the form. Other levels: 11 AAA, 12 AA, 13 A+, 14 A,
  16 Rookie, 17 Winter, 22 College, 23 Independent, 51 International. schedule.py
  reads it to pull the chosen levels. (51 carries WBC during the Classic, but also
  collegiate/Appalachian games at other times — documented as such.)
- Selecting International (51) adds the 20 WBC national teams to the team pickers;
  a national team already configured stays selectable even if 51 is later
  deselected, so it is never silently dropped. Only MLB + WBC teams are offered,
  since those are the only teams the board's color/ID maps resolve.

## Display (all layouts)

The awkward per-resolution coordinate page was replaced with a single
"Display (all layouts)" section exposing the four line_score booleans users
actually flip — show_hits_and_errors, show_abs_challenges, compress_digits,
shorten_team_name_on_high_line_score. Saving writes them into every
coordinates/*.json at once (surgical update of existing custom layouts; minimal
deep-merge override for sizes without one) so the setting holds regardless of the
panel resolution in use.


WardBrian:
I can review tomorrow but my first instinct is don’t we already have https://mlb-led-scoreboard.github.io/?

data/schedule.py
```diff
@@ -36,8 +36,8 @@ def update(self, force=False) -> UpdateStatus:
             LOGGER.debug("Updating schedule for %s", self.date)
             self.starttime = time.time()
             try:
-                # add sportId=51 to additionally get WBC games
-                all_games = statsapi.schedule(self.date.strftime("%Y-%m-%d"), sportId="1,51")
+                sport_ids = ",".join(str(s) for s in (self.config.sport_ids or [1]))
```
WardBrian: It's a small thing, but we could compute this string during config load to avoid repeatedly doing it

schemas/config.schema.json
```diff
@@ -404,17 +405,46 @@
       "minimum": 0
     },
     "api_refresh_rate": {
-      "description": "How often we update the status of a game in seconds",
+      "description": "How often we update the status of a game in seconds. Lower values call the MLB API more often.",
       "type": "number",
-      "default": 5,
-      "minimum": 1
+      "default": 15,
```
WardBrian, line 410: I think with the pitch clock we decided 15 was too infrequent

schemas/config.schema.json
```diff
@@ -404,17 +405,46 @@
...
+      "x-format": "range"
+    },
+    "sport_ids": {
+      "title": "Leagues",
+      "description": "Which leagues/levels to pull games from. MLB only by default. Note that 'International' (51) carries WBC games during the World Baseball Classic, but also collegiate/Appalachian League games at other times of year.",
+      "type": "array",
+      "items": {
+        "type": "number",
+        "enum": [1, 51, 11, 12, 13, 14, 16, 17, 22, 23],
+        "meta:enum": {
```
WardBrian, line 423: Question: should we just accept these strings and then convert to the number in the config class?

WardBrian (commented):
A few thoughts on the config -- I didn't review the web config code yet, which I think we should split off for separate discussion

tracerrx:
> I can review tomorrow but my first instinct is don’t we already have https://mlb-led-scoreboard.github.io/?

This runs locally on the RPI instead of on the web... I vote we keep it as it may be easier for some users (they wont have to SCP the config file over etc)...

WardBrian:
> This runs locally on the RPI instead of on the web... I vote we keep it as it may be easier for some users (they wont have to SCP the config file over etc)...

There are some nicer options we should consider, like giving a one-command-copy button: https://github.com/WardBrian/mlb-led-scoreboard/blob/318cd3e49dcb0a46d01b1cd20a4289802f49a03e/config.html#L115-L124

This never got ported to the live version of the site but could and would be much less new code

schemas/config.schema.json
ty-porter: League labels over magic numbers make sense to me.

ty-porter:
I don't have time to fully review this today, however before I do that I'd like to raise an alternative

As Brian mentioned we have https://mlb-led-scoreboard.github.io/ that already has a config editor. You can simply clone that repo and serve the static files via:

```sh
python -m http.server . <port>
```

Understandably this PR looks like it does a little bit more than just an editor, but it might be a smaller lift to implement there instead of inline with the main scoreboard.

twfarley:
> I don't have time to fully review this today, however before I do that I'd like to raise an alternative
> 
> As Brian mentioned we have https://mlb-led-scoreboard.github.io/ that already has a config editor. You can simply clone that repo and serve the static files via:
> 
> ```shell
> python -m http.server . <port>
> ```
> 
> Understandably this PR looks like it does a little bit more than just an editor, but it might be a smaller lift to implement there instead of inline with the main scoreboard.

Yeah, there's a bit of duplicated work for sure. But this lives on-device and edits current config.json without the scp back and forth. One less thing for users to work with. 

I'm happy to pull back the scope. It is a big change even with just support for the extra sportIds.

WardBrian:
I also think the solution to the sportIds being rather unspecific might be passing leagueIds as well, from https://statsapi.mlb.com/api/v1/leagues:

`sportId=1,51 leagueIDs=103,104,159,160`

Will show just MLB and WBC

README.md
```diff
@@ -265,6 +265,34 @@ See [RGBMatrixEmulator](https://github.com/ty-porter/RGBMatrixEmulator) for emul
...
+
+#### Web Config Editor
+
+Rather than hand-editing JSON, you can run a small local web app that presents every option as labeled checkboxes, dropdowns, and sliders. It's driven directly by the config schema, so the same descriptions you see below appear as inline help.
+
+Run it from the project root (in the same virtualenv the scoreboard uses):
+
+```sh
+# Reachable at http://<your-pi>.local/ on your network (port 80 needs sudo)
+sudo venv/bin/python config_editor.py
```
ty-porter, line 278: I would not recommend using `sudo` unless we absolutely have to. The scoreboard itself needs it for specific kernel features, editing some JSON definitely doesn't need it

config_editor.py
```diff
@@ -0,0 +1,483 @@
...
+EXAMPLE_CONFIG_FILE = ROOT / "config.example.json"
+CONFIG_SCHEMA_FILE = SCHEMAS_DIR / "config.schema.json"
+
+# Service name candidates probed for the optional "restart" button.
+SERVICE_NAME_PATTERN = re.compile(r"mlb.*scoreboard", re.IGNORECASE)
+
+# ── JSON helpers ────────────────────────────────────────────────────────────────
+
+
+def deep_update(base, overrides):
```
ty-porter, line 56: This is a duplicate of `bullpen.utils.deep_update`, any reason it can't be imported instead?

config_editor.py
```diff
@@ -0,0 +1,483 @@
...
+LINE_SCORE_KEYS = [
+    "show_hits_and_errors",
+    "show_abs_challenges",
+    "compress_digits",
+    "shorten_team_name_on_high_line_score",
+]
+PRIMARY_SIZE = "w128h64"
+
+
+def get_line_score():
```
ty-porter, line 255: So we only handle subsets of coordinates files?

ty-porter (commented):
I think where I'm stuck is I can't get over how much simpler it would be to modify what we have on the existing config editor to allow it to read/write local config files and serve it outside this project. This PR is introducing a schema resolver, an HTTP webserver, and tying both of those to scoreboard validators. The schema should guarantee we are making valid config changes without all the complexity.

Having written the config validator which does a similar thing (recursive validation, merging changesets), maintaining the schema resolver is going to be a huge PITA. I would strongly prefer a library that can handle that for us so we can focus on baseball things.

Security-wise I also think we should be agnostic on webserver, this allows you to expose an unsecured API with filesystem access (in some cases under `sudo`) to open internet, if you wanted.

twfarley:
Closing this — you were right that it was more surface area than the problem warranted.

Two things settled it. Upstream's `leagues` key (with WBC narrowed to `leagueId=159,160`) already covers what the `sport_ids` option here was reaching for, and does it better, so a good chunk of this PR is now redundant. And the static editor plus a one-command-copy button, as Brian suggested, gets most of the on-device benefit without a webserver and a schema resolver in the scoreboard repo.

I'll keep the editor on my fork since it's useful to me locally. Thanks both for taking the time to look at it.

PR #767, opened 2026-08-02
WBPL games via statsapi-like adaptor api

WardBrian:
This adds a new league option (#766) "WPBL"

image

image

---

The Women's Pro Baseball League has an api hosted at https://stats.womensprobaseballleague.com

Known endpoints:
- Schedule: `v1/games` https://stats.womensprobaseballleague.com/v1/games
- Game overview: `v1/games/{ID}` https://stats.womensprobaseballleague.com/v1/games/8alsgvzc90ypwphl
- Game play-by-play: `v1/games/{ID}/boxscore` https://stats.womensprobaseballleague.com/v1/games/8alsgvzc90ypwphl/boxscore
- Teams: `v1/teams` https://stats.womensprobaseballleague.com/v1/teams
- Team info: `v1/teams/{ID}` https://stats.womensprobaseballleague.com/v1/teams/v4gisr4rbgmn67b0/

This API is fairly different from `statsapi`. To get something basic off the ground and edit as little as possible, I wrote a small wpbl_statsapi_adaptor which queries these and returns objects that look more like what we expect from `statsapi`. There are definitely some missing values (e.g., I can't find a way to get the due up batters, or the winning pitcher), but it seems fine for basic display. 

Will need to monitor as a live game actually happens and update accordingly. We might also run into rate limiting issues with our default refresh rates.


data/game.py
```diff
@@ -68,11 +69,18 @@ def update(self, force=False, testing_params={}) -> UpdateStatus:
...
+                        {"gamePk": self.game_id, "fields": API_FIELDS} | testing_params,
+                        request_kwargs={"headers": data.headers.API_HEADERS},
+                    )
+
+                else:
+                    live_data = statsapi.get(
+                        "game",
+                        {"gamePk": self.game_id, "fields": API_FIELDS} | testing_params,
+                        request_kwargs={"headers": data.headers.API_HEADERS},
+                    )
```
ty-porter: Not opposed to this as-is to get beta support out for WPBL

I think these should be a true adapter pattern -- pass game ID to the top-level adapter which selects either statsapi or WPBL as the request backend. (I think this is what you were discussing in Discord, just calling it out explicitly).

Either way, this an easy refactor if we don't do this now.

ty-porter (approved):

data/game.py
WardBrian: I decided the right thing to do is associate each 'league' with its own API. Curious what you think of the latest push

PR #769, opened 2026-08-14
Add workflow to automatically reset dev after release

ty-porter:
I do this process by hand as part of a release, this just automates the happy path

There's still case where we commit directly to master, we'd have to handle that by hand. It's infrequent, so not very worried about it.

### DIFF
--- a/.github/workflows/update-dev.yml
+++ b/.github/workflows/update-dev.yml
@@ -1,0 +1,23 @@
+name: Update Development Branch on Release
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [master]
+
+permissions:
+  contents: write
+
+jobs:
+  update-dev:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true && github.event.pull_request.head.ref == 'dev'
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: master
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/update-dev.yml
```diff
@@ -0,0 +1,23 @@
+name: Update Development Branch on Release
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [master]
```
WardBrian, line 6: Can't just about anyone open a PR against master and then close it?

Do we also make a tag for releases? It seems saver to trigger anything like this off of the tag creation, since it can only be done with permissions

ty-porter: Yeah, I have a condition below it that it has to be merged + come from dev to try to constrain it specifically to our release PRs. Anything else could blow real stuff away including tags

with the goal being (1) don't merge a PR directly to master and (2) if you really need to, resolve the dev inconsistency by hand

WardBrian: I can’t remember the exact conditions github uses to pick which copy of a workflow to run if it’s edited in a PR, but there’s a world in which someone could edit that later condition out right?

ty-porter: It could, I don't think that's super high risk of happening, but I get the concern. If we have heartburn over the change we could close it, it's a minor optimization at best

PR #786, opened 2026-09-14
verbose 128x64 layout

twfarley:
New data from statsapi
  - Batter's season AVG / HR / RBI
  - Batter's spot in the batting order (plus on-deck and in-the-hole)
  - Pitcher's season ERA 
  - Play-by-play text: the resolved play, or the pitch just thrown in prose ("Andrew Sears throws a 94mph Four-Seam Fastball (Called Strike)")
  
  New optional layout elements (each draws nothing unless a layout names it)
  - atbat.batter_order — order number in place of the AB: label
  - atbat.batter_stats — AVG/HR/RBI, right-aligned to the panel edge; show_era adds the pitcher's ERA in place of P:
  - atbat.play_description — the play-by-play line; static when it fits, scrolls once when it doesn't 
  - final.record — season records beside the final score
  - inning.break.show_field — break screen keeps the diamond, outs and inning indicator (dimmed) instead of "Mid 5th"
  - inning.break.due_up.scroll — due-up batters on one scrolling line with order numbers
  - teams.abs_challenges.*.height — challenge marks as bars instead of squares
  - inning.arrow absolute x/y — both halves drawn, active bright, upcoming half blinks between innings
  
  New config option
  - layout_variant — selects an alternative layout file for your board size

  The layout itself
  - w128h64VERBOSE — teams and big scores in the bottom half, two rows of text up top, all of the above turned on. screenshots in the readme
  
  Bug fix (separate commit, droppable)
  - Challenge/review statuses now match case-insensitively — "Umpire Challenge" was skipping the shorthand and spilling 112px across a 128px panel


coordinates/README.md
```diff
@@ -16,6 +16,13 @@ You can edit these coordinates to display parts of the scoreboard in any way you
 2. Edit the coordinates in `w64h32.json` as you see fit
 3. Your customized coordinates will always take precedence over the example defaults
 
+## Alternative Layouts
+
+A file whose name carries a suffix after the dimensions, such as `w128h64VERBOSE.example.json`, is an alternative arrangement for that board size rather than a separate size. Load one with the `layout_variant` config option (`"layout_variant": "VERBOSE"`), and customize it by copying it to `w128h64VERBOSE.json` — the same `.json` override rule as any other layout.
+
+> [!IMPORTANT]
+> Do not rename a variant to the plain `w<cols>h<rows>` name. A custom file is reconciled against the example whose name it matches, and keys the example does not have are **deleted**, so a variant copied to `w128h64.json` would lose everything specific to it on the next update.
+
```
WardBrian, line 25: I am not in favor of `layout_variant` existing. I think we can either update the default 128x64 layout or not, and just leave these options for users to discover.

Alternatively, if we really want to include it, I think we should make the way you have to use it be by just copying it to `w128h64.json`, rather than introducing a new config option

renderers/games/game.py
```diff
@@ -86,53 +112,260 @@ def __render_play_result(canvas, layout, colors, play_result):
...
+
+def __render_play_description(canvas, layout, colors, description):
+    """Draw the play-by-play line: static if it fits, else scroll it once.
+
+    Short descriptions are common -- "Mound visit.", "Strikeout.", "Wild pitch." --
+    and scrolling one that already fits is just harder to read for no gain. Only
+    text wider than the slot animates, and it returns a non-zero width while doing
+    so, which holds the rotation until it has been read once.
+    """
+    global _play_desc_pos, _play_desc_last, _play_desc_finished
```
WardBrian, line 224: This seems like a real hack that will be annoying to keep in sync with the global scrolling state

renderers/games/game.py
```diff
@@ -381,3 +672,47 @@ def __render_inning_half(canvas, layout, colors, inning):
...
+        inactive = colors.graphics_color("inning.arrow.inactive")
+    except KeyError:
+        return
+
+    if status.is_inning_break(inning.state):
+        # Blink whichever half is about to be played, at 1 Hz. Middle means the
+        # bottom is next; End means the top of the next inning is.
+        #
+        # Wall clock rather than the renderer's animation_time, which only advances
+        # during a play animation and is pinned at 0 the rest of the time.
```
WardBrian, line 692: This is also a hack I'm not super happy with

tests/test_game_renderer.py
WardBrian, line 1: I think this kind of mock tests are super fussy and tend to break mostly when you're _intentionally_ changing behavior, making that more annoying. I wouldn't be opposed to something more like an expect/golden test using the emulator to create an image, since it is easy to just say 'accept the new image as correct now'

Ty does something like this in the [emulator](https://github.com/ty-porter/RGBMatrixEmulator/blob/main/test/reference.py#L152)

I also wouldn't mind just not having automated tests for this. It's pretty easy to run the board and see if the game renderer is doing what we intend

data/game.py
```diff
@@ -299,6 +312,76 @@ def pitcher(self):
...
+                try:
+                    order = self._current_data["liveData"]["boxscore"]["teams"][side]["players"][ID]["battingOrder"]
+                except (KeyError, TypeError):
+                    continue
+                return int(order) // 100
+        except (KeyError, TypeError, ValueError):
+            pass
+        return None
+
+    def __offense_id(self, slot):
```
WardBrian, line 352: This could also be used in the above functions for the name of the in hole batter, etc

data/game.py
```diff
@@ -299,6 +312,76 @@ def pitcher(self):
...
+        """Spot in the order for the current batter, or None."""
+        return self.batting_order_for(self.__offense_id("batter"))
+
+    def on_deck_batting_order(self):
+        return self.batting_order_for(self.__offense_id("onDeck"))
+
+    def in_hole_batting_order(self):
+        return self.batting_order_for(self.__offense_id("inHole"))
+
+    def pitcher_era(self):
```
WardBrian, line 368: Could this use `pitcher_stat` above? I'm pretty sure we already use that for era in pre/postgame

data/game.py
```diff
@@ -371,6 +454,86 @@ def current_play_result(self):
...
+        # the resolved play arrives seconds later and says what actually happened,
+        # so announcing "in play, out(s)" first is both clumsy and redundant.
+        if call and call.lower() != pitch.lower() and not call.lower().startswith("in play"):
+            sentence += f" ({call})"
+        return sentence
+
+    def pitcher_full_name(self):
+        try:
+            pitcher_id = self._current_data["liveData"]["linescore"]["defense"]["pitcher"]["id"]
+            return self.full_name(pitcher_id)
```
WardBrian, line 499: The only other places we use `full_name()` in the code base appear to also be for pitchers, so I feel like we could either use this elsewhere or not have this helper

data/game.py
```diff
@@ -36,6 +37,18 @@
...
+    """ "a" or "an" for a speed or a pitch name.
+
+    Spoken aloud, "88" starts with a vowel ("eighty-eight"), so "a 88mph Slider"
+    reads wrong on a line that is otherwise prose.
+    """
+    text = str(value)
+    if text[:1] == "8":
+        return "an"
+    return "an" if text[:1].lower() in "aeiou" else "a"
+
```
WardBrian, line 50: I'd prefer if this was at the bottom of the file or in a utility file elsewhere

data/game.py
```diff
@@ -371,6 +454,86 @@ def current_play_result(self):
...
+        Most specific first:
+
+        1. the resolved play, which MLB only populates once an at-bat ends;
+        2. the pitch just thrown, which refreshes every delivery and so covers
+           most of an at-bat;
+        3. the last resolved play we saw, held rather than cleared.
+
+        (3) can be stale -- across a pitching change or between innings there is
+        nothing live to say -- but holding the last thing that happened beats
+        leaving the line blank for minutes at a time.
```
WardBrian, line 515: Does it really?

I'd also note that this won't work well when rotating, since the Game objects get created/destroyed frequently

coordinates/README.md
ty-porter: Now might be a good time to introduce CLI flags to support `--coords` / `--colors` similarly to `--config`. But yes I otherwise agree -- default "off" and make things discoverable.

We used to have example configs for people to look at in the wiki, perhaps we could stand up a small repo with examples? With our schema validator it is easy to see if they're still usable in the current version or not

data/scoreboard/atbat.py
```diff
@@ -11,3 +11,10 @@ def __init__(self, game: "Game"):
         self.onDeck = game.on_deck()
         self.inHole = game.in_hole()
         self.pitcher = game.pitcher()
+        self.batting_order = game.batter_batting_order()
+        self.onDeck_order = game.on_deck_batting_order()
+        self.inHole_order = game.in_hole_batting_order()
```
ty-porter, line 16: Let's pick a consistent casing 😆

coordinates/README.md
WardBrian: That's a nice idea. People sometimes ask for mine, which is pretty different from the default 64x32

ty-porter (commented):
This looks really cool. Brian left better feedback than I could so I won't add more on top.

I would primarily be concerned with yet another way we handle configs (alternate layout machinery) as opposed to shipping 1 example layout + discoverable options. I think there are maybe better ways to handle the layout discovery. That doesn't necessarily need to be solved in this PR

Other than that, these new options look great

renderers/main.py
```diff
@@ -175,6 +181,17 @@ def __draw_plugin_screen(self, plugin_name: str, cond: Callable[[], bool]) -> No
 
         renderer.reset()
 
+    def __blink_on(self) -> bool:
+        """The on half of a 1Hz blink, for elements that pulse on their own schedule.
+
```
WardBrian: I still have the feeling that this should be the same as animation timing.

tests/test_final_records.py
WardBrian, line 1: Similar to my comment from the first review, I don't find this kind of mock based testing to be a value add

README.md
```diff
@@ -453,6 +454,38 @@ If you're feeling adventurous (and we highly encourage it!), the sections below
...
+* The pitcher's name and season ERA.
+* A line of play-by-play text — the resolved play when there is one, otherwise the pitch just thrown ("Andrew Sears throws a 94mph Four-Seam Fastball (Called Strike)").
+* Both halves of the inning indicator at once, the active half bright, so which half is being played reads the same way all game. Between innings the upcoming half blinks.
+* Team names and scores in the bottom half, at a size readable across a room, with the ABS challenge markers against the team colour edge.
+* Between innings, the diamond and out markers stay on screen dimmed and the due-up batters scroll along one line.
+* On the final screen, `FINAL` and the winning/losing pitcher line sit above the teams, with each team's season record beside its score.
+
+Every one of those is an individual coordinate option — see the `coordinates/` directory for what each one does and how to turn it on in a layout of your own.
+
+> The screenshots above are rendered from the test fixtures, not a live game.
```
WardBrian:
```suggestion

```

renderers/games/game.py
```diff
@@ -38,7 +40,26 @@ def render_live_game(canvas, layout: Layout, colors: Color, scoreboard: Scoreboa
         _render_outs(canvas, layout, colors, scoreboard.outs)
         _render_bases(canvas, layout, colors, scoreboard.bases, scoreboard.homerun(), (animation_time % 16) // 5)
 
-        _render_inning_display(canvas, layout, colors, scoreboard.inning)
+        # Optional full play-by-play line (no-op unless the layout enables it).
+        # Game.current_play_description() already picks the best available text --
+        # resolved play, else the pitch just thrown -- so there is nothing left for
+        # the renderer to substitute.
```
WardBrian:
```suggestion

```

renderers/games/game.py
```diff
@@ -38,7 +40,26 @@ def render_live_game(canvas, layout: Layout, colors: Color, scoreboard: Scoreboa
...
+
+    elif __break_shows_field(layout):
+        # An opt-in break screen that keeps the live screen's furniture instead of
+        # the "Mid 5th" text. The diamond and the out markers stay put but dimmed
+        # -- between halves of an inning there are no runners and no outs to
+        # report, so drawing them lit would state something false, while dropping
+        # them makes the board visibly lose half its content.
+        #
+        # The inning number and the blinking arrow then carry what the text used to
+        # say, in a fraction of the space.
```
WardBrian:
```suggestion

```

renderers/games/game.py
```diff
@@ -212,7 +412,10 @@ def _render_count(canvas, layout, colors, pitches: Pitches):
 
 
 # --------------- outs ---------------
-def __out_colors(colors):
+def __out_colors(colors, override_color=None):
+    if override_color is not None:
+        return [override_color] * 3, [override_color] * 3
```
WardBrian, line 438: All this override_color logic feels a little messy to all these functions. A better solution doesn't immediately come to mind, and I could probably live with it, just wanted to flag my instinct

renderers/games/postgame.py
```diff
@@ -20,9 +20,42 @@ def render_postgame(
...
+    """Season records beside the final score, when the layout asks for them.
+
+    Separate from `teams.record`, which the banner draws on every screen: on a
+    board where the live screen already fills the space to the right of the team
+    block, records can only go there once the game is over. Drawn from here rather
+    than from the banner so nothing else has to become screen-aware.
+
+    The record MLB reports in `gameData` for a completed game already includes that
+    game -- verified against the standings on a game that had just gone final -- so
+    there is nothing to wait for and no correction to apply.
```
WardBrian:
```suggestion

```

renderers/main.py
WardBrian: I still have the feeling that this should be the same as animation timing.

WardBrian (commented):
I'm overall pretty happy with this now! Most of my review is asking to remove some very verbose comments

renderers/main.py
twfarley: Agreed, done. MainRenderer keeps a free-running `frame_count` and the renderer derives the blink from it as `(frame_count // 6) % 2` -- same period and same units as the play-result animation right above it. The wall clock is gone.

It does have to be a second counter rather than `animation_time` itself: that one is deliberately pinned at 0 unless the current play is a homerun/strikeout/hit/walk, and `_render_at_bat` gates the play-result flash on it. Letting it run free would start flashing `error` and `fielders_choice`, which are in `PLAY_RESULTS` but not in that condition.

tests/test_final_records.py
twfarley: Deleted. Same conclusion as the renderer tests from the first round -- pinning this down with mocks costs more to maintain than it catches.

renderers/games/postgame.py
twfarley: Cut down to the one-line summary.

renderers/games/game.py
twfarley: Removed.

renderers/games/game.py
twfarley: Removed.

README.md
twfarley: Gone.

renderers/games/game.py
twfarley: Leaving it as-is for now -- nothing better came to mind here either. Both call sites want the same thing, one flat colour for every element on the dimmed break screen, so the kwarg is at least uniform. If a third case turns up I would rather the break renderer pass a palette down than keep widening each signature.

data/scoreboard/atbat.py
twfarley: Fixed -- `batter_order` / `on_deck_order` / `in_hole_order`, and the `onDeck` / `inHole` attributes that were already there went snake_case with them.

renderers/main.py
WardBrian: I think we could just change the condition in _render_at_bat to not use `results = list(PLAY_RESULTS.keys())`, and then we really could just use animation for everything?

twfarley: You are right, and it turned out to be hiding a bug that predates this branch.

`animation_time` now just runs free, `render_live_game` computes the phase once, and the same value drives the play-result flash and the arrow. The separate counter is gone, and `results = list(PLAY_RESULTS.keys())` with it -- `play_result in PLAY_RESULTS` reads better and stops rebuilding a list every frame.

The reason I thought the gate was load-bearing: `error` and `fielders_choice` are in `PLAY_RESULTS` but were not in the homerun/strikeout/hit/walk condition. So on any layout with `atbat.play_result` enabled -- five of the six stock ones -- those two took the play-result branch, which suppresses the batter name, while `animation` stayed pinned at 0 and never drew the `E`/`FC`. The at-bat row went blank for the whole play. Dropping the gate is what makes them render. `plays.OTHERS` is `[ERROR, FIELDERS_CHOICE]` and has never been referenced anywhere, which I assume is where that condition was supposed to come from.

I walked every `PLAY_RESULTS` key over a full flash cycle before and after; the only behaviour change is those two going from drawing nothing to drawing their text.

ty-porter (approved):
Good from my perspective. Maybe clear with @WardBrian before merging?

WardBrian:
I'm happy with how the code looks, I meant to pull the branch and try it locally (mainly just to make sure that everything still functions properly with all the new settings turned _off_)

twfarley:
I'm okay making this an off-season effort if you want to run with this for the season remainder. I'd hate to make any regressions in the postseason.

WardBrian:
That's fair enough

